PR #4, opened 2016-06-17
User

dotemacs:
Added devise with views
Added the user model

Will need to migrate to HAML/Slim from ERB


run.sh
```diff
@@ -0,0 +1,30 @@
...
+    bundle exec rake db:create
+    bundle exec rake db:migrate
+    bundle exec rake db:seed
+    ;;
+setup)
+    echo 'running setup'
+    bundle exec rake db:setup
+    ;;
+esac
+bundle exec puma -p 3000 -C ./config/puma.rb
```
filipposc5: replace 3000 with $PUMA_PORT

PR #11, opened 2016-08-01
Style search

aliuk2012:


app/views/correspondence/index.html.slim
```diff
@@ -1,15 +1,23 @@
...
 = form_tag '/search', method: 'get'
-  = label_tag "Search for: "
-  = text_field_tag :search
-  = submit_tag "Submit"
+  .form-group
+    = label_tag :search, class:'form-label' do
+      .label-text
+        = "Search"
+        span.form-hint
+          = 'For example correspondent name, email address'
```
EdwardAndress: we're only searching on name at the moment

PR #23, opened 2016-08-11
Do not overwrite category with blank

EdwardAndress:
- if the user submits the edit form with a blank option selected we should not be over-writing the respective field with an empty string
- instead, parse the params and do not update with blank values / empty strings
- there may be a neater way of doing this without an additional parse method


app/controllers/correspondence_controller.rb
```diff
@@ -35,6 +35,10 @@ def search
 
   private
 
+  def parsed_edit_params
+    edit_correspondence_params.delete_if { |_key, value| value == "" }
```
jsugarman: 👍 for delete_if and _key for unused args

spec/controllers/correspondence_controller_spec.rb
```diff
@@ -98,6 +98,12 @@
       it 'updates the correspondence record' do
         expect(Correspondence.first.topic).to eq 'courts'
       end
+
+      it 'does not overwrite entries with blanks (if the blank dropdown option is selected)' do
+        id = all_correspondence.first.id
+        patch :update, params: { id: id, correspondence: { category: '', topic: 'courts' } }
```
jsugarman: do u need the topic: 'courts'. would remove unless you are going to test it or is needed?!

jsugarman:
looks good, one thing to ensure is that you are not deleting from params directly (i do not believe you are) and an alternative method would be to use except
`params.except(:courts)`

spec/controllers/correspondence_controller_spec.rb
EdwardAndress: The params will always have the :topic key, so this test is mimicking reality

PR #24, opened 2016-08-15
Deadlines

EdwardAndress:
- new Category class
- seed for Category 'rake db:seed'
- type / category specific deadlines achieved through time limit attributes on Category
- time limit does not include the starting date
- therefore, a time limit of 9 actually means there are 10 business days available for the work to be done


db/seeds.rb
```diff
@@ -5,5 +5,6 @@
 #
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
-
-
+puts "----Seeding Categories----"
+Category.find_or_create_by!(name: 'freedom_of_information_request', abbreviation: 'foi', internal_time_limit: 9, external_time_limit: 19)
+Category.find_or_create_by!(name: 'general_enquiry', abbreviation: 'gq', internal_time_limit: 9, external_time_limit: 14)
```
aliuk2012: This should read 

`Category.find_or_create_by!(name: 'Freedom of information request', abbreviation: 'foi', internal_time_limit: 9, external_time_limit: 19)`

db/seeds.rb
```diff
@@ -5,5 +5,6 @@
 #
 #   movies = Movie.create([{ name: 'Star Wars' }, { name: 'Lord of the Rings' }])
 #   Character.create(name: 'Luke', movie: movies.first)
-
-
+puts "----Seeding Categories----"
+Category.find_or_create_by!(name: 'freedom_of_information_request', abbreviation: 'foi', internal_time_limit: 9, external_time_limit: 19)
```
aliuk2012: This should read 

`Category.find_or_create_by!(name: 'Freedom of information request', abbreviation: 'foi', internal_time_limit: 9, external_time_limit: 19)`

spec/factories/categories.rb
```diff
@@ -0,0 +1,17 @@
+FactoryGirl.define do
+  factory :category do
+
+    name "freedom_of_information_request"
```
aliuk2012: This should read 
`name "Freedom of information request`

spec/factories/categories.rb
```diff
@@ -0,0 +1,17 @@
...
+    abbreviation "foi"
+    internal_time_limit 9
+    external_time_limit 19
+
+    trait :gq do
+      name "general_enquiry"
```
aliuk2012: This should read 
`name "General enquiry`

spec/controllers/correspondence_controller_spec.rb
```diff
@@ -102,7 +102,7 @@
       it 'does not overwrite entries with blanks (if the blank dropdown option is selected)' do
         id = all_correspondence.first.id
         patch :update, params: { id: id, correspondence: { category: '', topic: 'courts' } }
-        expect(Correspondence.first.category).to eq 'freedom_of_information_reuqest'
+        expect(Correspondence.first.category.name).to eq 'freedom_of_information_request'
```
aliuk2012: I think this should read as :
`expect(Correspondence.first.category.name).to eq 'Freedom of information request'`

app/views/correspondence/edit.html.slim
```diff
@@ -7,7 +7,7 @@ section.question.content tabindex=0
   
   = form_for @correspondence do |f|
     .column-half
-      = f.collection_select :category, Settings.correspondence_categories, :underscore, :humanize, {selected: @correspondence.category, include_blank: "Please select an option"}
+      = f.collection_select :category, Category.all, :id, :name, {selected: @correspondence.category, include_blank: "Please select an option"}
```
aliuk2012: I think this needs to be changed to this 
`= f.collection_select :category, Category.all, :id, :name, {selected: @correspondence.category_id, include_blank: "Please select an option"}`

EdwardAndress: i think so too!

EdwardAndress: thanks

db/seeds.rb
EdwardAndress: why?

db/seeds.rb
EdwardAndress: why?

db/seeds.rb
aliuk2012: Shouldn't the name attribute not include underscores (i.e humanized)? Otherwise anywhere you call Category.name you will need to humanise the value. The reason I picked this up is because after seeding the data etc and editing a question I was getting this 
screen shot 2016-08-16 at 11 06 19

We could humanise the display value for the select list then there's other parts that rely on it too like show view. The risk is we end up having loads of 'humanise' in the view unless we have some sort of presenter for those value to clean up the views. For example in the show view we already have 3 values that are using humanise.

aliuk2012: screen shot 2016-08-16 at 11 11 51
We could end up with a scenario like this which automated tests might not catch because if it tests for either value it could return a false pass.

app/views/correspondence/edit.html.slim
aliuk2012: needed to use :category_id instead of :category for the attribute to be updated correctly.

db/seeds.rb
EdwardAndress: ok you win

misaka: Also, it would make more sense to use the "real" time limit numbers here, so instead of `internal_time_limit: 9, external_time_limit: 19` it should be`internal_time_limit: 10, external_time_limit: 20`. Suggestion for how to handle this below.

app/services/deadline_calculator.rb
```diff
@@ -0,0 +1,19 @@
+class DeadlineCalculator
+
+  class << self
+    def internal_deadline(correspondence)
+      correspondence.category.internal_time_limit.business_days.after(start_date)
```
misaka: We should do a calculation here to adjust the time limit. For example, when the internal time limit is 10, we shouldn't require the caller to set the time limit to 9 (as [commented in the seeds.rb change](https://github.com/ministryofjustice/correspondence_tool_staff/pull/24/files#diff-f7a25bc30a409fc77050703fc048ca8cR9)).

We can do this with something like:

```
adjusted_internal_time_limit = correspondence.category.internal_time_limit - 1
adjusted_internal_time_limit.business_days.after(start_date)
```

(^^^ code untested, just guessing 😄 )

config/business_time.yml
```diff
@@ -0,0 +1,13 @@
+business_time:
+  beginning_of_workday: 9:00 am
+  end_of_workday: 5:00 pm
+  holidays:
```
misaka: `holidays` is being set in `business_time.rb`, so do we can probably remove these from here and replace with a comment mentioning to look in `business_time.rb`.

spec/features/correspondence_deadlines_spec.rb
```diff
@@ -0,0 +1,76 @@
+require 'rails_helper'
+
+feature 'Deadlines:' do
+
+  background do
+    Timecop.freeze('21/08/2016') do
```
misaka: You could also avoid having to use `Timecop` if we had an attribute `requested_at` for `Correspondence`.

spec/features/correspondence_deadlines_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+
+    context 'for General Enquiries' do
+      scenario 'is shown in the list view' do
+        visit '/'
+        rows = all('.report tr').select { |row| row.text.include?('General enquiry') }
+        rows.each { |row| expect(row.text.include?('05/09/16')).to be true }
+      end
+
+      scenario 'is shown in the detail view' do
+        visit "/correspondence/#{Correspondence.last.id}"
```
misaka: At the risk to being nit-picky, maybe use a `let` block, or some other way of naming this correspondence, so that we're not dependant on the order of creation of the correspondence fixtures here.

spec/features/correspondence_deadlines_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+    end
+    login_as create(:user)
+  end
+
+  context 'The internal deadline' do
+
+    context 'for General Enquiries' do
+      scenario 'is shown in the list view' do
+        visit '/'
+        rows = all('.report tr').select { |row| row.text.include?('General enquiry') }
```
misaka: Just an idea, but it might be good to use [SitePrisms](https://github.com/natritmeyer/site_prism) to abstract the page structure out.

spec/features/correspondence_deadlines_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+        rows.each { |row| expect(row.text.include?('05/09/16')).to be true }
+      end
+
+      scenario 'is shown in the detail view' do
+        visit "/correspondence/#{Correspondence.last.id}"
+        expect(page).to have_content('Internal deadline')
+        expect(page).to have_content('05/09/16')
```
misaka: It's not quite clear what's being tested here. Maybe we can rework these a little bit to be clearer that we're testing that internal and external deadline are different between `gq` and `foi` requests?

spec/features/correspondence_deadlines_spec.rb
misaka: I say that since this means a DB migration we create a card and park this in the backlog. We need to get moving on the webform stuff for the time being. @EdwardAndress whatcha think?

EdwardAndress: Agreed

misaka:
👍

aliuk2012:
👍

PR #35, opened 2016-10-19
WIP: Ct 138

EdwardAndress:
https://dsdmoj.atlassian.net/browse/CT-138

As an Assigner of an FOI
I need to add a new request to the system
So a Caseworker can take responsibility and deal with the request

This ticket is to allow the user to enter data into the system, and for that data to be stored in the data base as an entry.

**I still need to remove the API and, also, the search field**  I propose to do these as separate pull requests.


app/views/correspondence/new.html.slim
```diff
@@ -0,0 +1,20 @@
+= render partial: 'layouts/header', locals: {page_heading: t('.new_case')}
+
+= form_for @correspondence do |f|
+  = f.radio_button_fieldset :topic, choices: Settings.correspondence_topics
```
EdwardAndress: I will do that in a separate commit ;)

app/views/correspondence/new.html.slim
```diff
@@ -0,0 +1,20 @@
...
+  = f.email_field :email
+
+  = f.email_field :email_confirmation
+
+  = f.text_area :postal_address
+
+  = f.text_area :message
+
+  .form-group
+  	= f.gov_uk_date_field :received_date, legend_text: 'Received date'
```
EdwardAndress: You're right!

spec/rails_helper.rb
```diff
@@ -1,3 +1,6 @@
+require "codeclimate-test-reporter"
+CodeClimate::TestReporter.start
+
```
EdwardAndress: There's a few commits on this branch which have been in since you pushed it up...  I don't know how they got in.  Any thoughts?

app/controllers/correspondence_controller.rb
```diff
@@ -22,8 +38,8 @@ def update
   end
 
   def assign
-    if @correspondence.update(assign_correspondence_params) && @correspondence.drafter
-      flash.now[:notice] = "Correspondence assigned to #{@correspondence.drafter.email}"
+    if @correspondence.update(assign_params) && @correspondence.drafter
+      flash.now[:notice] = "Case assigned to #{@correspondence.drafter.email}"
```
misaka: More content for i18n.

app/views/correspondence/index.html.slim
```diff
@@ -15,53 +15,28 @@
...
-        |
-      th scope='col'
-        | Type
-      th scope='col'
-        | Topic
-      th scope='col'
-        | Name
-      th scope='col'
-        | Received
+        | Requester
```
misaka: I18n?

app/views/correspondence/index.html.slim
```diff
@@ -1,4 +1,4 @@
-= render partial: 'layouts/header', locals: {page_heading: 'List of questions'}
+= render partial: 'layouts/header', locals: {page_heading: 'New cases'}
```
misaka: I18n?

db/migrate/20160722121207_create_correspondence.rb
```diff
@@ -4,7 +4,6 @@ def change
       t.string :name
       t.string :email
       t.string :category
-      t.string :topic
```
misaka: Why are we changing the `correspondence` migation here, but adding new `correspondence` migrations below? It should be one or the other.

run.sh
```diff
@@ -0,0 +1,15 @@
...
+create)
+    echo "running create"
+    bundle exec rails db:setup
+    ;;
+migrate)
+    echo "running migrate and seed"
+    bundle exec rails db:migrate
+    ;;
+esac
+bundle exec puma -d -C config/puma.rb
```
misaka: Take a look at the [latest `run.sh` in the public tool](https://github.com/ministryofjustice/correspondence_tool_public/blob/34c601ee826f085633d9f9eb428e5e1bb296d108/run.sh#L14) for the right way to invoke puma without having to do the tail below.

docker-compose.yml
```diff
@@ -0,0 +1,17 @@
...
+      dockerfile: Dockerfile.development
+    environment:
+      RAILS_ENV: development
+    command: bundle exec puma -b tcp://0.0.0.0:3000
+    volumes:
+      - .:/usr/src/app
+    ports:
+      - "3000:3000"
+    depends_on:
+      - db
```
misaka: newline! ;)

spec/controllers/correspondence_controller_spec.rb
```diff
@@ -2,36 +2,58 @@
...
+        received_date_yyyy: Time.zone.today.year.to_s
+      }
+    }
+  end
 
-  context "anonymous user" do
+  before { create(:category, :foi) }
 
-    context ' GET index' do
+  context "as an anonymous user" do
```
misaka: 👍 much better use of contexts here

spec/controllers/correspondence_controller_spec.rb
```diff
@@ -70,10 +94,28 @@
...
+
+      it 'renders the new template' do
+        expect(response).to render_template(:new)
+      end
+    end
+
+    describe 'POST create' do
+      context 'with valid params' do
+        it 'makes a DB entry' do
+          expect { post :create, params: valid_params }.
```
misaka: Just an idea ... but why not move `valid_params` closer to the code here, within the `context` block? That keeps the spec's parameters closer to the implementation ... and you could also call them `params` instead of `valid_params`. You might lose a little bit of semantic labeling but if they're defined within this context it should be evident anyway, and you don't create extra unnecessary variables. Just a thought.

config/locales/en.yml
```diff
@@ -28,7 +28,15 @@ en:
...
   helpers:
     label:
       correspondence:
         user_id: 'Drafter'
+        name: Full name
+        message: Full request
+        postal_address: Address
+  correspondence:
+    new:
+      new_case: New case
```
misaka: No end-of-line.

spec/features/02_create/correspondence_create_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    expect(page).to have_content('New case')
+    fill_in 'Full name', with: correspondence.name
+    fill_in 'Email', with: correspondence.email
+    fill_in 'Email confirmation', with: correspondence.email
+    fill_in 'Full request', with: correspondence.message
+    fill_in 'Day', with: '01'
+    fill_in 'Month', with: '01'
+    fill_in 'Year', with: '2016'
+    click_button 'Create'
+
```
misaka: Also, if I'm not mistaken, there are only two tests for creating a case, this here and the "POST create" test in `spec/controllers/api/v1/correspondence_controller_spec.rb` ... but neither tests that the actual values submitted are the values created in the DB. I'd expect that one or perhaps even both tests would test that the correct values are put into the DB for the provided inputs.

app/controllers/api/v1/correspondence_controller.rb
```diff
@@ -18,11 +18,10 @@ def create
       def correspondence_params
```
misaka: As discussed in chat, it'll be easier to cut these out in another branch.

PR #38, opened 2016-11-02
Add brakeman to Guardfile

EdwardAndress:
Guard runs automatically when we make PRs (and warnings block the merge).
Let's make guard run brakeman so we can spot these issues earlier


Gemfile
```diff
@@ -50,15 +50,16 @@ group :development, :test do
 end
 
 group :development do
-  # Access an IRB console on exception pages or by using <%= console %> anywhere in the code.
-  gem 'web-console'
+  gem 'guard-brakeman'
```
misaka: We don't for `guard-rubocop` ... is it really worth it? It's not like it'll actually break code.

EdwardAndress: OK chaps... someone decide on this :)

PR #40, opened 2016-11-07
Rake task to create dev users.

misaka:
I mentioned that I had added a task to create dev users ... here's the PR. Might be a good idea to test this locally before approving?

This change is a bit tricky to test, now that we've made the restriction that it only not run in the deployed prod environment, as one can't just do `RAILS_ENV=production ENV=prod be rake users:dev_entries` as that'll try to connect to the production DB.


aliuk2012:
TBH I think it would be better to implement something we can use on dev/staging and not just restrict to local.

misaka:
@aliuk2012 Sure, that makes sense. Looking at the environment variables available in the Docker containers there is one `ENV` which looks suitable.

misaka:
@aliuk2012 Thinking about this further, does it still make sense to use `drafter@localhost` if it's a non-local environment? Locally there's no requirement that these email addresses work, it's probably best if they don't in fact, but in deployed environments we may prefer to have working email addresses. Do we **need** this to work in the deployed environments now?

misaka:
Having spoken to @aliuk2012 ILR we've agreed that there's no need to have working emails at this time for dev and staging deployed environments, so we'll continue with the `@localhost` email addresses.

misaka:
@aliuk2012 Updated to only not run in production-prod environments.

PR #44, opened 2016-11-15
CT-194 Responsive table layouts

aliuk2012:
https://dsdmoj.atlassian.net/browse/CT-194

Includes
- List should be responsive
- List is accessible for screen readers

Does not include
- Full row should be a hyperlink including white space

See ticket comments about why Full row hyperlink is not included.


misaka:
Just an aside, the failure mode of putting the i18n code into the `aria-label`s is pretty ugly ... it tries to create HTML within the tag:

screen shot 2016-11-16 at 12 48 59

Probably not worth the effort to fix it, just noticed it because my Rails server needed to be restarted before it picked up the translations.

PR #46, opened 2016-11-18
Sort cases in index by external deadline then ID

EdwardAndress:
https://dsdmoj.atlassian.net/browse/CT-204

* Sort cases in the list view by target date (aka external deadline)
* Then by ID (to ensure consistent order for cases whose target dates are identical)


app/models/correspondence.rb
```diff
@@ -2,6 +2,10 @@ class Correspondence < ApplicationRecord
 
   acts_as_gov_uk_date :received_date
 
+  scope :by_deadline, lambda {
+    order("(properties ->> 'external_deadline')::timestamp with time zone ASC")
```
aliuk2012: No sign of sorting by ID or any other secondary sorting.

EdwardAndress:
@misaka are you happy with the changes I've made since your original approval?  If so, I'll squash the two commits into one, then proceed with the merge.

PR #48, opened 2016-11-28
Assignment model created and used in current flow

EdwardAndress:
https://dsdmoj.atlassian.net/browse/CT-262

- I have implemented basic assignment of new (non-trigger) cases to drafters.
- I have also moved the view that will be used for accepting / rejecting assignments into the path of assignments_controller

I think the following are next steps, that should be in a different PR:
- accept / reject assignment
- reassignment
- deletion of previous drafter assignment upon reassignment


EdwardAndress:
@aliuk2012 and @misaka - I was not able to take my normal approach of building from the outside in, starting with a scenario / feature test.

There is clearly more work to be done here in terms of completing the feature.  I am unsure how much of that work should be done as part of this PR.  Please comment.

db/migrate/20161125115930_create_assignments.rb
```diff
@@ -0,0 +1,14 @@
+class CreateAssignments < ActiveRecord::Migration[5.0]
+  def change
+    create_table :assignments do |t|
+      t.string :assignment_type
+      t.string :status, default: 'pending'
```
misaka: Looks like `ActiveRecord` has support for an `Enum` type as of 4.1. While it doesn't have built-in support for PostgreSQL's Enum type, it looks like we [can do it manually](https://www.sitepoint.com/enumerated-types-with-activerecord-and-postgresql/). That would mean doing something like this inside the migration:

```ruby
execute <<-SQL
  CREATE TYPE status AS ENUM ('pending', 'rejected', 'accepted')
SQL
```

app/models/assignment.rb
```diff
@@ -0,0 +1,21 @@
+class Assignment < ApplicationRecord
+
+  validates :assignment_type, :status, presence: true
+  validates :status, inclusion: %w[pending rejected accepted]
```
misaka: We can [use Rails's enumerated types](https://www.sitepoint.com/enumerated-types-with-activerecord-and-postgresql/). This would mean doing:

```ruby
  enum status: { pending: 0, rejected: 1, accepted: 2 }
```

See my comments on the migration.

app/models/assignment.rb
```diff
@@ -0,0 +1,21 @@
...
+  validates :assignment_type, inclusion: %w[to_casework to_draft]
```
misaka: Can we actually link the `assignment_type` to the roles we have for users? Assuming we have roles that translate directly to the types here, which I would've called `caseworker` and `drafter` respectively.

db/migrate/20161125115930_create_assignments.rb
```diff
@@ -0,0 +1,14 @@
+class CreateAssignments < ActiveRecord::Migration[5.0]
+  def change
+    create_table :assignments do |t|
+      t.string :assignment_type
```
misaka: See my comment about re-using roles in the model. If we change the `assignment_type` to refer to a role, this would be:

```ruby
  t.references :role
```

Or similar.

app/models/correspondence.rb
```diff
@@ -39,8 +40,8 @@ def self.search(term)
     where('lower(name) LIKE ?', "%#{term.downcase}%")
   end
 
-  def drafter
-    user
+  def current_drafter
+    assignments.select(&:open_draft?).last.assignee
```
misaka: This has been making me scratch my head a bit. I think the name maybe didn't do enough to communicate what it's doing, and now that I've parsed the code, I'm wondering if this is correct. What if the correspondence goes through this flow:

# gets assigned to responder1
# gets re-assigned to responder2
# gets re-assigned back to responder1

Would we then have two entries for responder1 in `assignments`? We'd have to for this methods to be accurate. I think it'd make more sense to examine the state transition history to see what the last responder assignment was ... of course, we don't have that implemented yet, but do we really need this? It's used in `CorrespondenceController` but arguably it's not needed there, and could be better implemented.

app/models/user.rb
```diff
@@ -5,6 +5,8 @@ class User < ApplicationRecord
     :recoverable, :trackable, :validatable
 
   has_many :correspondence
+  has_many :assignments_made, class_name: 'Assignment', foreign_key: 'assigner_id'
+  has_many :assignments_received, class_name: 'Assignment', foreign_key: 'assignee_id'
```
misaka: What do you think about just calling this `assignments`? It doesn't look too ambiguous to me.

app/views/correspondence/edit.html.slim
```diff
@@ -11,11 +11,6 @@ section.question.content tabindex=0
       br
       = f.submit 'Save', class:'button'
 
-    .column-half
```
EdwardAndress: removed until we know how reassignment will work

app/views/correspondence/show.html.slim
```diff
@@ -15,12 +15,6 @@ section.question.content tabindex=0
     
     = link_to 'Edit', edit_correspondence_path(@correspondence), class:'button'
 
-  = form_for @correspondence, url: { controller: 'correspondence', action: 'assign' } do |f|
```
EdwardAndress: removed because assignment will not be done from this page

spec/features/correspondence_assignation_spec.rb
```diff
@@ -1,18 +0,0 @@
...
-
-  scenario 'from the detail view' do
-    login_as create(:user)
-    correspondence = Correspondence.first
-    visit "correspondence/#{correspondence.id}"
-    page.find(:select, text: 'jane_doe@drafters-example.com').select('jane_doe@drafters-example.com')
-    click_button 'Assign'
-    expect(page).to have_content("Case assigned to #{User.first.email}")
-  end
-end
```
EdwardAndress: again, this has been removed since assignment is about to change drastically

db/migrate/20161125115930_create_assignments.rb
EdwardAndress: I understand that this decouples the table from the model but am unsure as to how important that is here.  I.e. When are we going to need to understand the table without the model being within easy reach?  The downside is that we would need to add new values to the Postgres 'status' type if we decide that we want to use a new Asssignment#status.

db/migrate/20161125115930_create_assignments.rb
EdwardAndress: Role is just a User attribute - wouldn't this create role_id column?

app/controllers/correspondence_controller.rb
```diff
@@ -16,7 +16,7 @@ def create
 
     if @correspondence.save
       flash[:notice] = t('.case_created')
-      redirect_to correspondence_index_path
+      redirect_to new_correspondence_assignment_path @correspondence
```
misaka: Since we're changing the flow for correspondence creation here, we may want to look at including one of the other changes that's meant to go with this, which is to change the text on the submit button on the new correspondence page from "Create" to "Continue". See the screenshot for "Copies the details from the email request into the system" on the [Design pages for Sprint 10](https://dsdmoj.atlassian.net/wiki/pages/viewpage.action?pageId=119112074) to see what I mean.

db/schema.rb
```diff
@@ -10,11 +10,25 @@
...
 
-ActiveRecord::Schema.define(version: 20161116153411) do
+ActiveRecord::Schema.define(version: 20161130164018) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
+  create_table "assignments", force: :cascade do |t|
+    t.string   "assignment_type"
+    t.integer  "state",             default: 0
```
misaka: So, small problem. Our enum isn't being dumped! Aaahhh... oh wait. I wonder ... hrm ... *furious-Googling* ... check this out: 

https://github.com/SchemaPlus/schema_plus_enums

Looks like this cleans up the migration code *and* adds support for dumping the enum into the schema. Shweet! Especially considering the other [alternative I'd found](https://gist.github.com/checkbutton/5d24ed4c6f5978caccbf).

spec/features/02_create/correspondence_create_spec.rb
```diff
@@ -31,7 +31,7 @@
     fill_in 'Year', with: Time.zone.today.year.to_s
     click_button 'Create'
 
-    expect(current_path).to eq correspondence_index_path
+    expect(current_path).to eq new_correspondence_assignment_path Correspondence.first
```
misaka: The use of `Correspondence.first` actually duplicates the use on line 37 below. Am I mistaken in thinking that this should, technically, be `Correspondence.last`? Whether we use `new_correspondence` or `Correspondence.last` here I don't know if I'm too fussed, but it'd be nice to be consistent.

app/views/assignments/edit.html.slim
```diff
@@ -11,4 +11,3 @@
     h2.form-label-bold = t('.request')
     .wrapper
       = @correspondence.message
```
misaka: I must be missing something here ... I can't see how this edit page provides a way to edit the assignment.

app/models/assignment.rb
```diff
@@ -0,0 +1,20 @@
+class Assignment < ApplicationRecord
+
+  validates :assignment_type, :state, presence: true
+  validates :assignment_type, inclusion: %w[to_casework to_draft]
```
misaka: Can the two above `validates` be merged into one?

Also, looks like `assigntment_type` could also be an enum, would it make sense to do that?

app/controllers/assignments_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+  before_action :set_assignment, only: [:edit, :update]
+  before_action :set_correspondence, only: [:new, :create, :edit]
+
+  def new
+    @assignment = @correspondence.assignments.new
+  end
+
+  def create
+    @assignment = @correspondence.assignments.new(
+      assignment_params.merge(assignment_type: 'to_draft')
```
misaka: The type of assignment should probably be set by the form being submitted, we will after-all be doing different types of assignments eventually, so shouldn't we implement that here?

db/migrate/20161125115930_create_assignments.rb
```diff
@@ -0,0 +1,26 @@
+class CreateAssignments < ActiveRecord::Migration[5.0]
+  def up
+    execute <<-SQL
+      CREATE TYPE state AS ENUM ('pending', 'rejected', 'accepted');
+    SQL
+
+    create_table :assignments do |t|
+      t.string :assignment_type
+      t.integer :state, default: 0, index: true
```
EdwardAndress: Actually, there were a few things not quite right here which were uncovered when I tried to address your comment - thanks!

- The enum hash was declared incorrectly in Assignment
- The default value should be 'pending', rather than 0
- Dropping the type should probably come after dropping the table, when rollingback (though both work, this is a better pattern I think)

db/migrate/20161125115930_create_assignments.rb
```diff
@@ -0,0 +1,26 @@
...
+      t.references :assigner
+
+      t.timestamps
+    end
+  end
+
+  def down
+    execute <<-SQL
+      DROP TYPE state;
+    SQL
```
EdwardAndress: Ha!  See above :)

app/views/assignments/edit.html.slim
EdwardAndress: This will be used to accept / reject, perhaps using a different name.  It's here for now because accepting / rejecting will update the `state` (and edit is the normal precursory action)

db/schema.rb
EdwardAndress: I've tried using SchemaPlusEnums but still no dump

EdwardAndress: I'll look into this tomorrow

spec/features/02_create/correspondence_create_spec.rb
EdwardAndress: I agree!  Changed it to use the variable

app/models/assignment.rb
EdwardAndress: Yes. I think it does make sense for it to be an enum too.

Since we are validating two attributes here, with different requirements, there's no way that I can think of which would end up in fewer lines.  We could have one `validates` per attribute:

`validates :state, presence: true`
`validates :assignment_type, presence: true, inclusion: <valid options>`

UNLESS... it's an enum, then we no longer need to validate for inclusion, I guess...?

app/controllers/assignments_controller.rb
EdwardAndress: I think we can leave it here for now and move it into the form when we want to start making assignments of different assignment_type ... but if you think that would be problematic I am happy to move it into the form now.

app/models/assignment.rb
misaka: I see what you mean about validating two attributes here. I think if we focus validations on the attributes, i.e. one per attribute as you've demonstrated, it's less complicated. It feels like your first alternative is an attempt at DRYing up code which, IMO, doesn't need it that badly since validations are already quite concise.

re: do we still need to validate inclusion for an enum ... it depends on whether Rails will give us useful errors without it (e.g. populate errors for the flash message). I'd assumed we still needed the validations to do that, but you're right ... maybe declaring an `enum` is enough. It might also be something that SchemaPlus Enums will give us.

EdwardAndress: The error message for invalid options here is not friendly.  In the app, however, we will be controlling this.  I.e. #assignment_type will be set in the backend. If we get that wrong, should the user see an error message that they can do nothing about or should the app just break?  Same for #state.

app/views/assignments/edit.html.slim
EdwardAndress: I've added a comment to the view so that we don't forget some controls for accept / reject are coming.

app/models/assignment.rb
misaka: Ya, good point, we're not going to be telling the user to fix their input so we don't, in fact, need friendly messages to be automatically generated for us.

db/schema.rb
EdwardAndress: I think we can fix this but it will require a little work (which I think should be put into the backlog and weighed up against our other tickets).

Or, I might try to contribute to the gem over Christmas!

Would it present a problem if we just left this as it is for the time being, with the above in mind?  Would it be better to take out the enums?

EdwardAndress: It's possible to do `rake db:structure:dump`, which generates a file called structure.sql then, after dropping and recreating the db, `rake db:structure:load` can be used in place of migration

At present, there are no failing tests or any other indication that not being able to dump the schema in the usual way has any ill effect.

EdwardAndress: OK, the CI tests on Travis were failing because we had specified db:schema:load in travis.yml - I've changed it to execute db:migrate instead and all is well...

spec/controllers/assignments_controller_spec.rb
```diff
@@ -0,0 +1,116 @@
+require 'rails_helper'
+
+RSpec.describe AssignmentsController, type: :controller do
+  before { create(:assignment) }
+  let(:create_params) do
+    {
+      correspondence_id: Correspondence.first.id,
+      assignment: { assignee_id: User.drafters.first.id }
```
misaka: Similarly to the above comment, is there a dependence on the creation of an `Assignment` that could be expressed here?

spec/controllers/assignments_controller_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+    }
+  end
+
+  context 'as an anonymous user' do
+
+    describe 'GET new' do
+      it 'redirects to sign in page' do
+        get :new, params: { id: Assignment.first.id, correspondence_id: Correspondence.first.id }
```
misaka: I don't think you want `id` here since you're creating a new assignment.

app/models/assignment.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  validates :assignment_type, :state, :correspondence, :assigner, :assignee,
+    presence: true
+
+  enum state: {
+    pending: 'pending', rejected: 'rejected', accepted: 'accepted'
+  }
+
+  enum assignment_type: {
+    caseworker: 'caseworker', responder: 'responder'
```
misaka: Apparently we don't know that this should be `responder` vs `drafter`. Until we have a clear answer to that, and seeing as we've already started using the name `drafter`, how about we stick with that?

spec/controllers/assignments_controller_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+    end
+  end
+
+  context 'as an authenticated user' do
+    before { sign_in create(:user) }
+
+    describe 'GET new' do
+      it 'renders the page for assignment' do
+        get :new, params: {
+          id: Assignment.first.id, correspondence_id: Correspondence.first.id
```
misaka: Similar to the anonymous user, we don't need `id` here as we're creating a new assignment.

spec/controllers/assignments_controller_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+      end
+
+      it 'does not create a new assignment' do
+        expect { post :create, params: create_params }.
+          not_to change { Correspondence.first.assignments.count }
+      end
+    end
+
+    describe 'GET edit' do
+      it 'renders the page for accept / reject assignment' do
```
misaka: This should probably be "redirects to sign-in page".

spec/controllers/assignments_controller_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+            correspondence_id: Correspondence.first.id,
+            assignment: { state: 'accepted' }
+          }
+      end
+
+      it 'does not update state' do
+        expect(Assignment.first.state).to eq 'pending'
+      end
+
+      it 'redirects to correspondence detail page' do
```
misaka: This should probably be "redirects to sign-in page".

spec/controllers/assignments_controller_spec.rb
```diff
@@ -0,0 +1,116 @@
+require 'rails_helper'
+
+RSpec.describe AssignmentsController, type: :controller do
+  before { create(:assignment) }
+  let(:create_params) do
+    {
+      correspondence_id: Correspondence.first.id,
```
EdwardAndress: I agree this could read better.  Hopefully the changes I'm about to push will sort this out.

spec/controllers/assignments_controller_spec.rb
EdwardAndress: as above

spec/controllers/assignments_controller_spec.rb
EdwardAndress: correct!

app/models/assignment.rb
EdwardAndress: Yep.  Good point.  Otherwise this could get very confusing.  Changed to drafter, consistently accross the app

spec/controllers/assignments_controller_spec.rb
EdwardAndress: oh dear, I was asleep

spec/controllers/assignments_controller_spec.rb
EdwardAndress: and again

spec/models/assignment_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'rails_helper'
+
+RSpec.describe Assignment, type: :model do
```
misaka: We should probably also be testing that the enum `state` and `assignment_type` are defined correctly.

spec/models/assignment_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  subject { build(:assignment) }
+
+  describe 'callbacks' do
+    describe '#update_correspondence' do
+      it 'is called after create' do
+        expect(subject).to receive(:update_correspondence)
+        subject.save
+      end
+
+      it 'changes the state of correspondence to awaiting_responder' do
```
misaka: 👍

EdwardAndress:
I think this could be DONE :)

spec/support/matchers/have_enum.rb
```diff
@@ -0,0 +1,14 @@
...
+    enums = subject.defined_enums.keys.map(&:to_sym)
+    enums.include?(name)
+  end
+
+  chain :with_values do |expected_values|
+    actual_values = subject.defined_enums[name.to_s]
+    actual_model_values = actual_values.keys
+    actual_db_values = actual_values.values
+
+    (actual_model_values == expected_values) && (actual_db_values == expected_values)
```
misaka: This test enforces a specific ordering, e.g. if the `assignment_type` enum is defined as `['caseworker', 'drafter']`, testing it against `['drafter', 'caseworker']` will fail. Do we want that?

spec/support/matchers/have_enum.rb
```diff
@@ -0,0 +1,14 @@
+RSpec::Matchers.define :have_enum do |name|
+  match do
+    enums = subject.defined_enums.keys.map(&:to_sym)
+    enums.include?(name)
```
EdwardAndress: maybe I am missing something but... name is a symbol already

spec/support/matchers/have_enum.rb
EdwardAndress: No, we do not!

app/views/assignments/new.html.slim
```diff
@@ -0,0 +1,3 @@
+= form_for [ @correspondence, @assignment ] do |f|
+	= f.collection_select :assignee_id, User.drafters, :id, :email
+	= f.submit
```
aliuk2012: The submit button should have a class of 'button' on it.

EdwardAndress: Done it :)

EdwardAndress:
@aliuk2012 and @misaka - I think I have dealt with the changes you requested.

EdwardAndress:
Can we merge this?

PR #49, opened 2016-12-08
CT-217 Create Ping.json check

aliuk2012:
https://dsdmoj.atlassian.net/browse/CT-217

ping.json is needed for pingdom to check if the service is up I had to update the config gem so that the config.rb initializer could match the one used in Ask Tool. The rest is pretty much a mirror
of how we implemented this in the Ask tool.


config/routes.rb
```diff
@@ -1,5 +1,7 @@
 Rails.application.routes.draw do
 
+  get 'heart_beat/ping'
```
misaka: This should probably be:

```ruby
get 'heartbeat/ping'
```

But, do we really need it since we have `/ping` routed below?

spec/controllers/heartbeat_controller_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'rails_helper'
+
+RSpec.describe HeartbeatController, type: :controller do
+
+  describe 'ping and heartbeat do not force ssl' do
```
misaka: Instead of stating the expectation here, we should be stating what we're testing ... I would expect to see the test in this one in the `describe '#ping'` below.

config/routes.rb
aliuk2012: You are right this shouldn't have been there. Fixed it

spec/controllers/heartbeat_controller_spec.rb
aliuk2012: Could you check this now?

spec/controllers/heartbeat_controller_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'rails_helper'
+
+RSpec.describe HeartbeatController, type: :controller do
+
+  describe '#ping' do
+    describe 'does not force ssl' do
```
EdwardAndress: Maybe I am being dumb but, I don't quite get this... how is a response code of 301 associated with SSL?

misaka: What I was trying to suggest here is that `describe` block should stick to tell us the class/method/bit of code that's being speced, while the `it` block should specify specific behaviour. Something more akin to this:

```ruby
describe '#ping' do
  it 'does not force ssl' do
    allow(Rails).to receive(:env).and_return(double(development?: false, production?: true))
    get :ping
    expect(response.status).not_to eq(301)
  end

  it 'returns JSON with app information' do
...
```

The other thing, having thought about this for a tiny bit, is shouldn't this policy of redirecting HTTP to HTTPS be handled by NGinx? Is it actually be handled in the Rails app somewhere?

aliuk2012: Ah I see, good point there is no need to have a "before" statement for only one test.

I'm not sure about your NGINX question but I think that's a really good thought about where's the redirect happening.

aliuk2012: 301 is suggesting there's a permanent redirect, i.e I use 'http://....' it then redirects to 'https://...'

aliuk2012: This got me thinking are we forcing ssl on both Ask and Answer Tool Apps? I'm not sure if this is being done at NGINX level but I've seen another service use [force_ssl](http://api.rubyonrails.org/classes/ActionController/ForceSSL/ClassMethods.html) in their application controller(https://github.com/ministryofjustice/advocate-defence-payments/blob/develop/app/controllers/application_controller.rb#L8)

Shouldn't we be doing something similar ?

aliuk2012:
@EdwardAndress @misaka I think I've addressed all your concerns. Could you review?

aliuk2012:
@EdwardAndress review/approval?

PR #53, opened 2016-12-14
Clean up production logging.

misaka:


EdwardAndress:
How will this help logging?

misaka:
@EdwardAndress Previously, puma was being run in the background, which meant all we would see in the logs was this:

```
Puma starting in single mode...
* Version 3.6.0 (ruby 2.3.1-p112), codename: Sleepy Sunday Serenity
* Min threads: 5, max threads: 5
* Environment: production
* Daemonizing...
(Nothing has been logged yet.)
```

So now puma is running in the foreground so that it will log things to STDOUT.

One thing that was missing from this commit was changing the STDOUT log formatting to be logstash-compatible, and re-enabling the logs to `production.log`.

EdwardAndress:
Cool.  Thanks for explaining.

PR #60, opened 2016-12-19
CT-310 Update the assign page

aliuk2012:
https://dsdmoj.atlassian.net/browse/CT-310

- Removed green flash notice on the new assignment view
- Added "Assign case" page heading
- Copy change: 'Create assignment' to 'Create and assign case'
- Added a 'Successfully created case' flash message on case list view


aliuk2012:
@misaka @EdwardAndress  I think we should review and merge this and then tweak the rest of assign page later.

aliuk2012:
I've split this ticket into two separate JIRA tickets and so the understanding work thats blocking this PR/work has been moved into ct-353.

app/views/assignments/new.html.slim
```diff
@@ -1,3 +1,7 @@
+= render partial: 'layouts/header', locals: {page_heading: t('.new_assignment')}
+
 = form_for [ @case, @assignment ] do |f|
-	= f.collection_select :assignee_id, User.drafters, :id, :email
-	= f.submit class: 'button'
+
+  -# f.radio_button_fieldset :assignee_id, choices: User.drafters.tap(&:email).to_sym
```
misaka: Do we want this commented-out line?

aliuk2012: Fixed it

PR #61, opened 2016-12-19
CT-316  Removed "Forgot your password?" link on sign in page

aliuk2012:
https://dsdmoj.atlassian.net/browse/CT-316

- Removed the forgotten password link for MVP


app/views/devise/sessions/new.html.slim
```diff
@@ -10,5 +10,4 @@
 
   = f.submit "Sign in", class: "button"
 
-= render "devise/shared/links"
```
misaka: The alternative to doing it this way is to remove `:recoverable` from the `devise` module list in the `User` module. Wouldn't that be a bit cleaner in case we want the other Devise shared link?

PR #62, opened 2016-12-20
CT-327 Update case list view copy

aliuk2012:
https://dsdmoj.atlassian.net/browse/CT-327

- Copy changes
  - 'My cases' to 'Cases'
  - 'Name' to 'Requester name'
  - 'Target' to 'External deadline'

- Refactored the site prism's css selectors used for tests (see commit message)
- Remove 'Draft/Internal deadline' from the views


app/views/cases/_fixed_details.html.slim
```diff
@@ -20,12 +20,6 @@ h2.visually-hidden
     span#escalation_deadline
       = l @case.escalation_deadline, format: :deadline
 
-  - if @case.requires_approval?
-    span.form-label
-      = t('.internal_deadline')
-    span#internal_deadline
-      = l @case.internal_deadline, format: :deadline
-
```
misaka: Not trying to be pedantic, but the original story was about removing these from the case list view, not all the views. Is this just doing work that's in other stories?

Also, if we're removing the internal deadline, should we be cleaning it up in the backend? DB, model, specs, etc?

app/views/cases/index.html.slim
```diff
@@ -19,21 +19,17 @@
...
         th scope='col'
-          = t('.internal_deadline')
-        th scope='col'
           = t('.external_deadline')
         th scope='col'
           = t('.status')
       tbody
         - @cases.each do |c|
           tr.case_row
-            td.name aria-label="#{t('.requester')}"
```
misaka: 👍

app/views/cases/_fixed_details.html.slim
misaka: No, I see now. Sounds like there should've/could've been a ticket about removing the internal_deadline from the front-end separate from the case list view. So this makes sense.

PR #63, opened 2016-12-20
Add model and route annotations.

misaka:
Ed mentioned this a while back and I agreed it'd be a nice to have.


aliuk2012:
I remember using this on another project. It is really good.

PR #69, opened 2017-01-11
ct-326 Generate case reference number

EdwardAndress:
- generate number using the format:

YYMMDD001 for the first case on a given day, then increment with each new case

- reset suffix to 001 on the next day
- done when:
Case number is generated using GMT and stored in the DB

- There is another ticket (CT-329) to surface this reference number in the case list view


app/models/case.rb
```diff
@@ -29,6 +29,7 @@ class Case < ApplicationRecord
   validates :email, format: { with: /\A.+@.+\z/ }, if: -> { email.present? }
   validates :postal_address, presence: true, on: :create, if: -> { email.blank? }
   validates :subject, length: { maximum: 80 }
+  #validates :reference, presence: true
```
aliuk2012: Why is this commented out? Shouldn't the case number be a required attribute?

spec/models/case_spec.rb
```diff
@@ -50,6 +50,7 @@
     it { should validate_presence_of(:message)       }
     it { should validate_presence_of(:received_date) }
     it { should validate_presence_of(:subject)       }
+    #it { should validate_presence_of(:reference)     }
```
aliuk2012: Why is this commented out? Shouldn't the case number be a required attribute?

app/models/case.rb
EdwardAndress: The case reference is generated automatically by a callback so, unless we change that in error, it will always be present on anything that goes into the DB.

There is a test that ensures the callback is used and case reference numbers are generated on cases, when they're saved.

Similarly, we don't validate the presence of deadlines or the trigger flag (on trigger cases).

spec/models/case_spec.rb
EdwardAndress: I just forgot to remove the comment because my normal practice is to delete rather than comment out.

db/structure.sql
```diff
@@ -3,10 +3,11 @@
 --
 
 -- Dumped from database version 9.5.5
--- Dumped by pg_dump version 9.5.5
+-- Dumped by pg_dump version 9.6.1
```
aliuk2012: I dont think our production db is using 9.6

EdwardAndress: I don't know if there's a problem with using different versions of pg_dump locally and remotely.  Given that we're not currently storing anything important in the DB and have no real users, we could do a test.

Also, when the dump is performed on the DB, the server will use whatever version of pg_dump it has installed.  So structure.sql will be generated appropriately, in theory.  If there is a problem with version mismatch, it should only be felt if we deploy the app and try to use structure.sql to set up the database, rather than running the migrations...

Those are my speculative thoughts.

EdwardAndress: I've tested this branch by deploying to dev, using both migrate and reset params.  All works out fine.  @aliuk2012 has also gone into the machine and ensured the case number is generated as expected. That is fine too.

PR #70, opened 2017-01-13
CT-329 Add case reference to list view

EdwardAndress:
CT-329

- add the new case reference number to the list view
- ticket does not mention creating this as a link to the detail view
- neither does it mention adding the number to the detail view


spec/support/pages/case_list_page.rb
```diff
@@ -1,4 +1,5 @@
 class CaseList < SitePrism::Section
+  element :reference, 'td[aria-label="Case number"]'
```
misaka: Just to check, should we be calling it 'case number' or 'case reference'? Just because we seem to be calling it 'reference' in the changes we did in the backend (e.g. #73) but 'number' here for the frontend ...

EdwardAndress: It's number in the designs, so that's what the user sees.  I preferred 'reference'.  In retrospect this may have been a little self indulgent.

aliuk2012: I think case reference for the back-end is self explainatory. I also dont think we should tie the backend attributes to closely to what is just copy on the pages.

1. What happens if the 'case number' isn't a number and becomes alphanumeric?
2. What if they decide to change the copy would we then rename all the backend attributes.

Could we have a chat about it today?

aliuk2012: see https://github.com/ministryofjustice/correspondence_tool_staff/pull/73

app/models/case_number_counter.rb
```diff
@@ -1,6 +1,6 @@
 # == Schema Information
 #
-# Table name: case_reference_counters
+# Table name: case_number_counters
```
misaka: How did this get in here? Wouldn't this have come from the case number PR?

PR #73, opened 2017-01-17
CT-326 case num format

misaka:
Some changes to the case reference numbers:

- use case_reference_counters table to track counters for case references
- change references to a string just for future compatibility
- add unique constraint on references, clean up tests to reflect this


db/structure.sql
```diff
@@ -3,11 +3,10 @@
 --
 
 -- Dumped from database version 9.5.5
--- Dumped by pg_dump version 9.6.1
+-- Dumped by pg_dump version 9.5.5
```
aliuk2012: Have we sorted this version issue out?

EdwardAndress: yep! :) We had to force my machine to use 9.5

db/migrate/20170111161049_add_reference_to_cases.rb
```diff
@@ -1,5 +1,6 @@
 class AddReferenceToCases < ActiveRecord::Migration[5.0]
   def change
-    add_column :cases, :reference, :integer, null: false
+    add_column :cases, :reference, :string, null: false
+    add_index :cases, :reference, unique: true
```
aliuk2012: We shouldn't reuse existing migrations otherwise we have to tear down the DB. Can you create a new migration?

aliuk2012:
I'm surprised you could use "number" as an attribute, I thought it had to be some kind of reserved word.

Apart from it might be a reserved word somewhere down the line( i.e case_type) case reference makes more sense to me:
1. Case number is not a "number but a string"
2. what happens when it doesn't contain just numbers?

misaka:
@aliuk2012 While both @EdwardAndress and I agree that 'reference' makes more sense than 'number', the problem is that we've been calling it 'number' in the frontend, the designs, etc. And like the 'correspondence' vs 'case' issue we had, this risks causing confusion when we work with it. I think that right now it may make more sense to us to use 'reference', but when the system starts to develop more, whatever term we use when discussing the product with everyone else, is going to be the one that'll make the most sense.

aliuk2012:
fine we'll leave it at that then.

PR #78, opened 2017-01-20
Ct 278 roles and permissions

EdwardAndress:
For CT-278

Implement permissions using either Pundit, CanCan, or another gem.

- I chose to do a spike with Pundit

https://github.com/elabs/pundit

- To this end, I created a simple feature test, where only assigners are able to view the New Case form.
- This works quite nicely and we can add custom flash messages using locales
- Pundit also comes with it's own rspec matchers as you will see!
- In addition, we will be able to use Pundit to return specific scopes on the index action, should we decide to show different users different list views, based on their role (although I suspect we are likely to go somewhat beyond that).
- There is also a method to ensure authorisation is checked after each action, should we decide to use it.


EdwardAndress:
If Pundit provides all the functionality we need and the syntax is not objectionable I vote we go with it @misaka and @aliuk2012

aliuk2012:
I suppose you could also test to ensure that Assigners cannot access the assignments edit page

EdwardAndress:
@aliuk2012 - sure, we could do that and I am happy to stick it in if we feel it's appropriate.  The idea, though, was to just add the simplest possible example of a 'policy' at this stage.

The policy for that edit page might be that you can only see the edit page for assignments where you are the assignee...

aliuk2012:
Good point about it would be restricted to drafters and the team its assigned for.

app/controllers/application_controller.rb
```diff
@@ -1,12 +1,26 @@
...
   def set_user
     @user = current_user
   end
 
+  private
+
+  def user_not_authorized(exception)
+    policy_name = exception.policy.class.to_s.underscore
+
+    flash[:error] = t "pundit.#{policy_name}.#{exception.query}"
```
aliuk2012: You need to replace ':error' with ':alert' to get the error message styled properly

EdwardAndress: Done!

aliuk2012:
😠  Flicker again - Subject text from factory girl longer than 80chars

https://travis-ci.org/ministryofjustice/correspondence_tool_staff/builds/195920336

app/views/cases/index.html.slim
```diff
@@ -1,8 +1,9 @@
 = render partial: 'layouts/header', locals: {page_heading: t('.page_heading')}
 
-.grid-row
-  .column-full
-    = link_to t('common.new_case_button'), new_case_path, class: 'button'
+- if policy(:case).new?
```
misaka: As discussed in person, I think that `new?` doesn't communicate very well what we're checking for here, `create_new?` or something similar would make this a little bit clearer.

app/policies/case_policy.rb
```diff
@@ -0,0 +1,13 @@
+class CasePolicy
+
+  attr_reader :user, :case
+
+  def initialize(user, kase)
+    @user = user
+    @case = kase
+  end
+
+  def new?
```
misaka: Shouldn't we have a `create?` here too? Should these be separate methods? Paired with the comment that `new?` in the view could be better communicated, one idea is to have `create_new?` in the policy, and then to alias `new?` and `create?` to it, to DRY things up.

EdwardAndress:
@misaka I have made some changes as per the chat we had earlier.

@aliuk2012 you may want to check out the second commit here

PR #79, opened 2017-01-20
CT-364 Install psql and migrate db

misaka:
We need postgresql-client so that we can easily connect to the DB console. And running migrations for every run is relatively cheap, when compared to not running migrations and breaking an environment. We also want to get this before we're properly in "production", so we can test that it doesn't have undesirable side-effects.


run.sh
```diff
@@ -10,9 +10,7 @@ create)
     echo "running create"
     bundle exec rails db:setup
     ;;
-migrate)
-    echo "running migrate"
-    bundle exec rails db:migrate
-    ;;
 esac
+
+bundle exec rails db:migrate
```
aliuk2012: Do we always want to run migrates here? Couldn't we just default the 'Migrate' setting on jenkins to 'migrate instead of 'none'

misaka: We could, but I'd be tempted to try to keep the wording so that running a migration is standard. Looking at the DOCKER_STATE param in Jenkins, it's already a little out-of-sync as there's a `reset` state that isn't handled in `run.sh`, what if we changed it so that the states are:

- regular startup -- run the migration and start up the app
- no-migration -- just start the app with no migration
- create -- create the db and start the app
- reset -- reset the db and start the app

aliuk2012: I like the sound of those states. Go for it, I just thought that maybe there would be a need to deploy without migrating and by not having a condition seemed a bit forceful.

However shouldn't the 'create' do migrate and seed?

aliuk2012: there is a comment

Dockerfile
```diff
@@ -4,7 +4,8 @@ ENV PUMA_PORT 3000
 
 RUN touch /etc/inittab
 
-RUN apt-get update && apt-get install -y
+RUN apt-get update && apt-get install -y \
+  postgresql-client
```
filipposc5: in general we install postgresql-client from pg repos directly as they provide a higher level of accuracy and faster updates for the different versions. Sample below may need adjustment.


This is a phusion/baseimage example I have handy:
```
a152m:pg9.4example filippos$ cat Dockerfile
FROM phusion/baseimage:0.9.16


RUN apt-get update -y  && \
    apt-get install wget ca-certificates  && \
    sh -c 'echo "deb http://apt.postgresql.org/pub/repos/apt/ $(lsb_release -cs)-pgdg main" > /etc/apt/sources.list.d/pgdg.list'   &&\
    wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | sudo apt-key add -  && \
    sudo apt-get update && \
    sudo apt-get install -y postgresql-9.4-postgis-2.3-scripts

```

run.sh
filipposc5: A small comment, changing those tasks makes jenkins jobs more "special" so I would suggest sticking to the known values and map them to the behaviour that you want. 

Eg: default state is none. (usually indicates no "task")
seed (usually create & seed?)
reset (usually drop & create & seed) 
migrate (run the migration)
finally always: start up the app

misaka: @filipposc5 I assume you mean want to keep the state names and meanings consistent. So what if we just set the default state to `migrate`, as @aliuk2012 originally suggested? Our goal is to always migrate, and as we said in HipChat we don't think that there will be a problem with race conditions.

Dockerfile
```diff
@@ -4,7 +4,12 @@ ENV PUMA_PORT 3000
 
 RUN touch /etc/inittab
 
-RUN apt-get update && apt-get install -y
+RUN apt-get update -y  && \
+    apt-get install wget ca-certificates  && \
+    sh -c 'echo "deb http://apt.postgresql.org/pub/repos/apt/ $(lsb_release -cs)-pgdg main" > /etc/apt/sources.list.d/pgdg.list' && \
```
misaka: `lsb_release` is failing when I run them by hand inside the existing containers. @filipposc5 is this going to be an issue when building the container? I'll try to find some time a little later to build the container myself to test, but I'm not sure I have everything I need installed.

Dockerfile
```diff
@@ -4,7 +4,12 @@ ENV PUMA_PORT 3000
...
+    wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | sudo apt-key add -  && \
+    sudo apt-get update && \
```
misaka: Also, we don't have `sudo` on these containers, but we probably don't need to run `sudo` anyway.

Dockerfile
misaka: Fixed by installing `lsb-release` package.

Dockerfile
misaka: Fixed by removing `sudo` from commands.

misaka:
This latest version has been tested by creating the Docker image locally.

EdwardAndress:
Two thumbs up

PR #81, opened 2017-01-24
CT-362 add sidekiq

misaka:
https://dsdmoj.atlassian.net/browse/CT-362

This adds an ActiveJob for our mail process using SideKiq and Redis. Basically, so that we can email out notifications to our users.

This work will be require changes in the deploy repo to make it to the deployed environments.


EdwardAndress:
Why is an empty job and, corresponding, empty spec added here?

misaka:
Because I wanted to have some kind of job in here, even if it gets renamed later, etc. I was assuming that we would require _a_ job to be able to spin up SideKiq.

EdwardAndress:
Can you test that?

EdwardAndress:
I approve either way

EdwardAndress:
Although given that ActiveJob is also compatible with other queueing systems, it would be possible to switch out Sidekiq for something else at some point in the future.  This would be made easier if adding sidekiq were separated from adding the jobs, at least in terms of the commits used.

misaka:
@EdwardAndress I didn't think it was worth proving, tbh. I don't think there's enough benefit to be had, even if we do decide to rip out SideKiq I don't think we'd do a revert commit, I think we'd just add the gem and edit the config files.

EdwardAndress:
OK

misaka:
@aliuk2012 @EdwardAndress Sorry for more changes coming through, but there was some things that needed fixing. If you could re-review I'd appreciate it. :)

PR #84, opened 2017-01-27
CT-363 Accept/Reject User Journeys

aliuk2012:
https://dsdmoj.atlassian.net/browse/CT-363

- If a Drafter rejects a case then take them to 'rejected' page and show them success message and read-only version of the case details

- If a Drafter accepts a case then take them to the case details page and show a success message

- If another user tries to access the original assignment after the case has been rejected take them to the 'rejected' page and tell them it has already been rejected

- If another user tries to access the original assignment after the case has been accepted then take them to the case detail page but do not show them any success message


app/controllers/assignments_controller.rb
```diff
@@ -48,6 +62,18 @@ def assignment_params
 
   def set_assignment
     @assignment = Assignment.find(params[:id])
+  rescue ActiveRecord::RecordNotFound
```
misaka: I can kind of see why codeclimate is complaining about this ... it'd be nicer to protect using `Assignment.exists?(params[:id])`.

app/views/cases/show.html.slim
```diff
@@ -1,3 +1,7 @@
+- if @accepted_now == 'true'
+  .alert-green
+    = t('alerts.successfully_assigned_html', case_list_url: cases_path)
```
EdwardAndress: Should we make the accept and reject message keys consistent?

I.e. If we're going to use 'alerts.successfully_assigned_html' should we also use 'alerts.successfully_rejected_html' etc. ?

Also, I think a message with 'assigned' is more relevant to the assigner than the responder, who accepts the assignment.  So, something more like '.successfully_accepted_html', except there is no unsuccessful acceptance, so perhaps that's misleading...

app/controllers/assignments_controller.rb
```diff
@@ -21,21 +21,35 @@ def create
...
-  def edit; end
+  def edit
+    if @assignment
+      if already_accepted?
+        redirect_to case_path @case, accepted_now: false
+      else
+        render :edit
+      end
+    elsif @assignment.nil? && already_rejected?
+      redirect_to case_assignments_rejected_path @case, rejected_now: false
```
misaka: Why isn't this a redirect to `case_path` like the above? Looking at the views below they look basically they same ... they both render the `case_detail_page` ... it seems we could just re-use that with a tiny bit more logic in it? Or would the case `#show` template be too complicated you reckon?

app/views/cases/show.html.slim
misaka: If I understand you correctly, maybe `.assignment_accepted_html` and `.assignment_rejected_html` would make sense?

app/controllers/assignments_controller.rb
misaka: Also, just to add to that, I initially thought that a `#rejected` action in a controller was a bit odd, since it's not actually an action. Another reason why it might make sense to rely on the case `#show`.

app/controllers/assignments_controller.rb
```diff
@@ -47,7 +61,20 @@ def assignment_params
...
+    if Assignment.exists?(id: params[:id])
+      @assignment = Assignment.find(params[:id])
+    end
+  end
+
+  def already_rejected?
+    rejection = @case.transitions.detect do |transition|
+      transition.assignment_id == params[:id].to_i && transition.event =='reject_responder_assignment'
+    end
+    rejection.present?
```
aliuk2012: 👍  fixed

app/views/cases/show.html.slim
aliuk2012: @EdwardAndress good point about keeping the message keys consistent. 

@misaka Those key names make sense. i've not also renamed the `. already_rejected_html` key to `.assignment_already_rejected_html' and placed it under "alerts"

app/controllers/assignments_controller.rb
aliuk2012: Assignment rejection is kind of like the end of the user journey for drafters so we should try and ensure that they cannot do anything else on the case. I did wonder about reusing `#show` but the problem with that is there is already a bit of logic in the views and I suspect there will be more added to it once we start laying out different case views for each type of user.

Do you think renaming the `#rejected` action might help clear things up?

aliuk2012: @misaka could you review/approve?

aliuk2012:
@misaka I've renamed the `#rejected` action to `#show_rejected`. Could you review?

app/controllers/assignments_controller.rb
```diff
@@ -21,21 +21,36 @@ def create
     end
   end
 
-  def edit; end
+  def edit
+    if @assignment
+      if already_accepted?
+        redirect_to case_path @case, accepted_now: false
```
misaka: Just a thought ... I wouldn't say to change this now, but Rails provides a cleaner-looking way of passing state between requests ... using the `flash`. The [Rail's ActionController overview](http://guides.rubyonrails.org/action_controller_overview.html#the-flash) mentions this as does [this blog posting on The Pug Automatic](https://thepugautomatic.com/2012/08/the-rails-flash/), so it's not that esoteric. Basically, you could do:

```
flash[:accepted_now] = true
redirect_to case_path @case
```

And then simply access it in the `CasesController`:

```
@accepted_now = flash[:accepted_now]
```

Avoiding exposing the state in the param.

spec/controllers/assignments_controller_spec.rb
```diff
@@ -147,15 +149,15 @@
           }
         end
 
-        it 'calls #reject' do
+        it 'calls #show_rejected' do
```
misaka: This probably should've stayed as `#reject`. Over-zealous search-replace? 😄

spec/controllers/assignments_controller_spec.rb
```diff
@@ -147,15 +149,15 @@
...
           allow(drafter_assignment).to receive(:reject)
           patch :accept_or_reject, params: update_params
           expect(drafter_assignment).to have_received(:reject).with(message)
         end
 
-        it 'redirects to case detail page' do
+        it 'redirects rejected assignments page' do
```
misaka: Looks like this is the test that should've been tweaked to `show_rejected`.

PR #89, opened 2017-02-03
Ct-409 assign/reassign unassigned cases

EdwardAndress:
- We need a mechanism that allows assignment of unassigned cases, outside the current case creation flow
- A button has been placed in the case detail page to achieve this
- It is only visible to assigners AND when the case is in unassigned state
- Some of the flash notices have also been changed to reflect that:
** At present case creation occurs after the initial New Case form is submitted
** Case assignment occurs via the assignment form


spec/features/02_create/case_create_spec.rb
```diff
@@ -61,9 +62,9 @@
     )
 
     select drafter.email, from: 'assignment[assignee_id]'
-    click_button 'Create and assign case'
+    click_button 'Assign case'
```
aliuk2012: Couldn't we use site prism?

EdwardAndress:
@aliuk2012 - yes, you're right about the policy.

EdwardAndress:
@aliuk2012 - I have, I think, addressed your comments regarding the policy :)

PR #90, opened 2017-02-03
Fix case_transition reference to case#transitions

EdwardAndress:
The relation is called #transitions, not #case_transitions


aliuk2012:
Looks alright but my only comment is whether theres a test or something that we could use to make sure this doesn't happen?

misaka:
👍 ... I was just going to comment about how this is boiler-plate code that was added by the gem, but good work getting it covered with a test @EdwardAndress :)

PR #91, opened 2017-02-03
Ct-408 KILO only see their assigned cases in list

EdwardAndress:
- KILOs should only see the cases which have been assigned to them in their case list view


app/policies/case_policy.rb
```diff
@@ -14,4 +14,24 @@ def can_add_case?
...
+    def initialize(user, scope)
+      @user  = user
+      @scope = scope
+    end
+
+    def resolve
+      if user.assigner?
+        scope.all
+      elsif user.drafter?
+        scope.select {|kase| kase.drafter == user }
```
aliuk2012: Shouldn't we should all case which have been assigned to a particular kilo? it might cause confusion whose doing what if we dont have "whose it with" status

EdwardAndress: This will show all the cases that have been assigned to the current user.

EdwardAndress: Does that answer your question?

EdwardAndress: I am confused

aliuk2012: Sorry you are right, I misunderstood what was happening here. A Kilo can only see their cases in the case list view. When I was manually testing it I was able to access another kilo's case by editing the url. I think we should maybe look to lock this down either as part of this PR or perhaps get it ticketed up as another piece of work

EdwardAndress: Yep!  Good point.  Also, if it's not clear we should try to improve that :)

aliuk2012: No I think its clear, I just need to review more carefully ;P

PR #99, opened 2017-02-13
Ct 421 mark response as sent

EdwardAndress:
- When there are attachments of type 'response' the assigned KILO should be able to 'Mark response as sent'
- This triggers a transition to 'responded' state
- I have also tweaked the logic around the 'Accept or reject' button to ensure the user will be taken to the correct assignment, even when we add case worker assignments in the future.

## This has been branched from my other work CT-411 so until that one is merged, there will be a bigger diff here than you might expect.


EdwardAndress:
Stuart has agreed that we should not allow assigners to upload documents for the MVP.  In the short term we have too many unanswered questions about how that would work.  The MVP, therefore, is restricted to a simple flow where assigners assign and drafters draft.

config/locales/en.yml
```diff
@@ -117,6 +117,12 @@ en:
       This case is being reviewed for sensitivity.
       Don't send out the response until <strong>%{escalation_deadline}</strong>.
     response_upload_error: Errors detected with uploaded files.
+    respond_point_of_no_return_html: |
+      <strong>Important</strong><br>
+      <p>
+        After this step you'll no longer be able to upload an updated<br>
```
misaka: Why the `<br>`? This is one sentence ...

EdwardAndress: It's the pea that proves you are a princess.

spec/features/04_respond_to_correspondence/mark_response_as_sent_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+        href: Rails.root.join("/cases/#{kase.id}"))
+    expect(case_list_page.case_list.first.number).
+      to have_link(another_kase.number,
+        href: Rails.root.join("/cases/#{another_kase.id}"))
+    expect(case_list_page).to have_content(
+      'Response confirmed. The case is now with DACU.'
+    )
+  end
+
+end
```
aliuk2012: I think these tests could be written to use site_prism else its pointless to have `spec/support/pages/case_response_page.rb`

app/views/cases/respond.html.slim
```diff
@@ -0,0 +1,9 @@
+= render partial: 'layouts/case_page_heading', locals: { case_details: @case }
+
+.reminders
+	= t('.reminders_html')
+
+.alert-orange
+	= t('alerts.respond_point_of_no_return_html')
+
+= link_to t('common.case.respond'), confirm_respond_case_path(@case), class: 'button', method: :patch
```
aliuk2012: Could you wrap the wrap into grid-rows and columns?

``` html
.grid-row
  .column-two-thirds
    .reminders
       .....

.grid-row
  .column-two-thirds
    .alert-orange
      ......

.grid-row
  .column-two-thirds
     = link_to .....
```

spec/features/04_respond_to_correspondence/mark_response_as_sent_spec.rb
EdwardAndress: All good points!  I have attempted to address them.

app/views/cases/respond.html.slim
```diff
@@ -0,0 +1,15 @@
+= render partial: 'layouts/case_page_heading', locals: { case_details: @case }
+
+.grid-row
+  .column-two-thirds
+    .reminders
+      = t('.reminders_html')
+
+.grid-row
+  .column-two-thirds
+		.alert-orange
```
misaka: What's happened with the indentation here? I guess slim doesn't actually care so maybe it doesn't matter.

spec/features/04_respond_to_correspondence/mark_response_as_sent_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+      )
+    expect(response_page).to have_link('Mark response as sent',
+                                  href: confirm_respond_case_path(kase))
+    click_link 'Mark response as sent'
+
+    expect(kase.current_state).to eq 'responded'
+    expect(current_path).to eq '/cases'
+    expect(case_list_page.case_list.count).to eq 1
+    expect(case_list_page.case_list.first.number).
+      not_to have_link(another_kase.number,
```
misaka: I think this should be `s/another_kase.id/kase.id/` ?

app/views/cases/respond.html.slim
EdwardAndress: Something odd.  That's not how it appears in my editor.

misaka: You're probably using tabs then. There have been many a battle about tabs vs spaces, but if it's making it harder to read in GItHub we may want to stick with spaces. The only  alternative, I believe, would be to find a way to define the tabstop for this file so that GitHub and other editors can pick them up.

misaka: http://stackoverflow.com/questions/8833953/how-to-change-tab-size-on-github/23522945#23522945 ... I'd prefer just to stick with spaces if everyone is ok with that.

EdwardAndress:
@misaka - you were correct about the spaces vs tabs thing.  I had switched to spaces earlier but only after having indented a few times with tabs.  Since it all looked OK I assumed that it was all OK... ERROR!

I've also addressed your other comment about the link in the case list.  Well spotted!!

app/views/cases/respond.html.slim
EdwardAndress: sure, I have no preference

spec/features/04_respond_to_correspondence/mark_response_as_sent_spec.rb
```diff
@@ -52,6 +52,26 @@
...
+    expect(responded_case_row.detail_link['href']).to eq case_path(kase.id)
+    expect(responded_case_row.subject.text).to eq kase.subject
+    expect(responded_case_row.name.text).to eq kase.name
+    expect(responded_case_row.status.text).to eq 'Waiting to be closed'
+
+    expect(drafting_case_row.detail_link).to have_content(another_kase.number)
+    expect(drafting_case_row.detail_link['href']).to eq case_path(another_kase.id)
+    expect(drafting_case_row.subject.text).to eq another_kase.subject
+    expect(drafting_case_row.name.text).to eq another_kase.name
+    expect(drafting_case_row.status.text).to eq 'Waiting for draft response'
```
misaka: This is getting mega-long. How about putting this new bit into a new scenario? It might require using factories, but this block is a good 50 lines long now. :-/

EdwardAndress: Or I could cut down on the expectations in that last section, since the idea is just to check that the correct cases are listed.

aliuk2012: Do we need to test all the columns? We just want to make sure the case is shown/hidden to the right user.

EdwardAndress:
OK, I've removed some bits and now it's no longer the longest scenario we have.

spec/features/04_respond_to_correspondence/mark_response_as_sent_spec.rb
```diff
@@ -41,17 +41,25 @@
...
-    expect(remaining_case_row.detail_link['href']).to eq case_path(another_kase.id)
-    expect(remaining_case_row.subject.text).to eq another_kase.subject
-    expect(remaining_case_row.name.text).to eq another_kase.name
+    expect(remaining_case_row.number).to have_content(another_kase.number)
     expect(remaining_case_row.status.text).to eq 'Waiting for draft response'
 
     expect(case_list_page).
       to have_content('Response confirmed. The case is now with DACU.')
+
+    login_as assigner
```
misaka: Ah yes, I was looking at this last night and didn't comment because I wanted to sleep on it. This still feels too long to me. :( It just feels like unless logging in as another user is part of the journey here somehow (assigner has to approve it, or something, and even in then ...), it's a bit too much. It feels to me like we're testing too much in here.

misaka: Is the reason you want to have this all in one scenario because it's a pain to setup the system for the assigner test?

EdwardAndress: Not so much that it's a pain but I want to see that the exact same case, which the drafter just marked as sent, is present in the assigner's case list.  Maybe that's not required, given the fact that it never left the assigner's case list view anyway...

misaka: You specifically wanted to have the same case number? Or did you just want to make sure that the system is in the same state?

If it's the former, then why?

If it's the later, then ya ... I can kind of see how it might be a pain, but my feeling is that that's a slipper slope and if we're never going to trust our factories to do a good enough job, then a lot of our tests are already invalid.

misaka: You could also just trim this down even further so this is less about testing what the assigner see, and very specifically that the assigner sees the case that disappeared for the drafter:

```ruby
login_as assigner
case_list_page.load
expect(case_list_page.case_list).to have_content(kase.number)
```

PR #100, opened 2017-02-13
Ct 352 - Add Service Feedback

aliuk2012:
https://dsdmoj.atlassian.net/browse/CT-352

- Adds Feedback model
- Add JS to to AJAX to submit the feedback form without submitting the whole page
- Setup Poltergeist for AJAX/JS Testing
- Allow shared DB Connections for tests. (Please check https://github.com/ministryofjustice/correspondence_tool_staff/pull/100/commits/658f19cb892ab62d9bbd9a2c5d225c16ce809e46#diff-f2369014861328398fb9a81dd6426247 as mentioned in the answer to this question http://stackoverflow.com/questions/18623661/why-is-capybara-discarding-my-session-after-one-event)


spec/features/00_feedback/submit_feedback_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    expect(home_page).to have_service_feedback
+    expect(home_page.service_feedback).to have_feedback_form
+    expect(home_page.service_feedback).to have_send_button
+
+    home_page.service_feedback.feedback_textarea.set "Very good service"
+    home_page.service_feedback.send_button.click
+
+    home_page.service_feedback.wait_for_success_notice
+    expect(home_page.service_feedback).to have_success_notice
+  end
```
EdwardAndress: What happens to the feedback box when feedback is submitted?  Is the text cleared?

EdwardAndress: Should we include some expectations which check that the feedback was saved?

spec/controllers/feedback_controller_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      post :create, params: params, xhr: true
+    }
+
+    it "return 200 if trying to submit new feedback using AJAX" do
+      expect(response.code).to eq "200"
+    end
+
+    it 'makes a DB entry' do
+      expect(Feedback.first.comment).to eq "This is my feedback"
+    end
```
EdwardAndress: Should we check the other attributes too?

spec/support/pages/feedback_section.rb
```diff
@@ -0,0 +1,13 @@
+class ServiceFeedback < SitePrism::Section
+  element :feedback_form, '#new_feedback'
+  element :feedback_textarea, '#new_feedback textarea'
+
+  element :send_button, '.button-secondary'
```
EdwardAndress: As an aside, maybe the upload documents button should be a secondary button, because it doesn't cause a state transition...

app/controllers/feedback_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  private
+
+  def create_feedback_params
+    params.require(:feedback).permit(
+        :comment,
+    ).merge({ email: current_user.email,
+              user_agent: request.user_agent,
+              referer: request.referer
+            })
```
EdwardAndress: Do we test anywhere that the 'referer' and  user_agent are set correctly?  Are there any circumstances in which these might be nil?  Will the application handle that?

misaka: I'm not sure that we can test that these are set correctly ... it's in-built Rails functionality that relies on headers in the HTTP request. We could try to catch the edge-case where they're set to `nil` to make sure it doesn't break the user journey, but then, if they're `nil` then there's probably something abnormal going on anyway (the user is actually a script, for example).

app/controllers/feedback_controller.rb
```diff
@@ -0,0 +1,26 @@
+class FeedbackController < ApplicationController
+
+  def create
+    @feedback = Feedback.new(create_feedback_params)
+
+    respond_to do |format|
+      if @feedback.save
```
misaka: Why an `if` statement here if the behaviour doesn't change?

app/assets/javascripts/modules/Feedback.js
```diff
@@ -0,0 +1,7 @@
+moj.Modules.Feedback = {
+  init: function(){
```
misaka: How about putting this logic into the `create.js.erb` below to keep the response logic the same?

app/views/feedback/_alerts.html.slim
```diff
@@ -0,0 +1,12 @@
+- if feedback.errors.any?
+  .alert-red
+    | The following error(s) occurred:
```
misaka: This line looks like a candidate for I18n.

spec/controllers/feedback_controller_spec.rb
```diff
@@ -0,0 +1,47 @@
+require 'rails_helper'
+
+RSpec.describe FeedbackController, type: :controller do
+
+  let(:params)do
+    { feedback: {
+        comment: "This is my feedback",
+        user_agent: 'firefox'
+    }
+    }
```
misaka: I know this is getting picky, but having two braces like this makes me immediately think that there's been a cut'n'paste error. I know we're not being terribly consistent with our code style here, which is ok, but could we at least avoid false-alarms like this? I would probably bust it out so that `feedback: {` is on a separate line, and then the first brace should line-up with that.

spec/controllers/feedback_controller_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  context "as an anonymous user" do
+
+    describe '#create' do
```
misaka: Could we move `context` and `describe` around so that instead of having:

```
context 'as an anonymous user' do
  describe '#create' do
```

```
describe '#create' do
  context 'as an anonymous user' do
```

So we can group together tests around *what* we're testing, instead of the various contexts we'll have?

app/controllers/feedback_controller.rb
aliuk2012: refer and user_agents aren't required attributes so if they are nil, then so be it they will just be recorded as nil. It just saves us having to contact the user to ask them things like "What page were you on and what browser were you using"

spec/models/feedback_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'rails_helper'
+
+RSpec.describe Feedback, type: :model do
+
+  let(:feedback) { build :feedback}
+
+  describe 'has a factory' do
```
misaka: Is this testing that FactoryGirl works? Isn't this overkill?

spec/features/00_feedback/submit_feedback_spec.rb
aliuk2012: 👍

spec/controllers/feedback_controller_spec.rb
aliuk2012: I've added emails as we need to know who it has come from.

spec/support/pages/feedback_section.rb
aliuk2012: Trick question, Upload button is primary if they have not uploaded any previous responses but once they have I think that switches to a secondary action as the next stage is for them to mark the case as responded

spec/controllers/feedback_controller_spec.rb
aliuk2012: 👍

spec/models/feedback_spec.rb
aliuk2012: I don't think this is overkill. All the other models spec have similar syntax but I think its good to make sure that models have a valid factory to use and test with

aliuk2012:
@EdwardAndress @misaka I think I've addressed all your comments. Could you review?

PR #102, opened 2017-02-15
CT-352 Feedback textarea not clearing

aliuk2012:
create.js.erb was using the wrong condition to
check if the feedback was successfully saved or
not and so it would never pass the js that clears
the textarea. The jquery object selector was also
missing a single quote which would have broken the
js.


EdwardAndress:
Should we have a test that covers this functionality?

aliuk2012:
@EdwardAndress @misaka oddly there is a test but it didn't catch it https://github.com/ministryofjustice/correspondence_tool_staff/blob/develop/spec/features/00_feedback/submit_feedback_spec.rb#L43

I'm going to merge this now as I don't think we should speak any more time on the feedback feature because we will be redoing a feedback.

PR #103, opened 2017-02-17
CT-417 add uploaded files to case detail view

misaka:
https://dsdmoj.atlassian.net/browse/CT-417

This PR moves the table of uploaded files to the case detail view, adding a Download button so users can view the attached file. The Download button isn't specified in the original ticket, but I thought it would make sense to have it.

Some of the commits here don't have the CT-417 prefix because I thought they weren't wholly and clearly unrelated to the ticket, they were just nice things to add as I went.


EdwardAndress:
- Does the download button actually download the file, to the user's computer, or just show the file in the browser?
- Do we need a policy around who can download these files?

misaka:
@EdwardAndress --

1. It should, have you tried it?
2. I've asked that very question to Stu in the JIRA ticket.

misaka:
@EdwardAndress - Just realised that my answer to 1. was potentially ambiguous ... it will start a download of the file to the computer. We may add in-app viewing later, but that's a bit of work so maybe we should figure out if it's needed first.

EdwardAndress:
Thanks @misaka - in that case, is it enough to test that the redirect occurs (because the rest, i.e. actually downloading the file, is not our code)?  Is there anything that could go wrong here as a consequence of our code?

misaka:
@EdwardAndress So in theory we could/may want to test that the content-type is appropriate, or something, but this gets a bit trickier. Do we actually want to make the request to S3 and see what the headers look like, for example? The problem there would be that this is the test bucket, so we wouldn't be testing the "right thing".

I would think that if we need to do any more work when adding attachments in the first place, like ensuring the mime type is correct, etc, we'd want to test that part as opposed to testing the effects that'll have against S3 itself.

EdwardAndress:
OK, makes sense!  Thanks @misaka :)

misaka:
@EdwardAndress I've decided to add further URL validation here, could you have a look? It's a separate commit so should be quick to double-check.

spec/models/case_attachment_spec.rb
```diff
@@ -42,17 +42,21 @@
 
     it { should validate_presence_of :url }
 
-    xit 'validates that the URL points to our storage bucket'
+    it 'validates that the URL points to our storage bucket' do
```
EdwardAndress: Does this correspond to the `validate_url_is_valid_host` validation?

misaka: Ya ... you're going to make a comment about naming, aren't ya. Kind of like `correspondence_response` ... ?

spec/factories/case_attachments.rb
```diff
@@ -12,16 +12,19 @@
...
     }
 
-    factory :correspondence_response do
+    factory :correspondence_response, parent: :case_attachment do
       type 'response'
     end
-
+    factory :case_response, parent: :correspondence_response do
+      # Whatever was I thinking calling it :correspondence_response? Why didn't
+      # you stop me Eddie?!?!
```
EdwardAndress: STOP @misaka

app/views/layouts/_case_detail_page.html.slim
```diff
@@ -15,6 +15,26 @@
             = t('common.case.received_date')
           = l case_details.received_date, format: :default
 
+      - if include_uploaded_files
```
aliuk2012: I think there should also be a condition to only show this if there are uploaded files otherwise the page looks broken.
image

misaka: Would you suggest to hide it, or to have copy saying that there's nothing uploaded? Something like:

_No files have been uploaded._

aliuk2012: I would just hide it by not including it for the time being. This also gives them an impression that they are "building" on to the case by uploading and performing actions.

app/views/layouts/_case_detail_page.html.slim
```diff
@@ -15,6 +15,26 @@
...
+        .request--uploaded-files
+          table#uploaded-files
+            colgroup
+              col
```
aliuk2012: missing another `col`

app/views/layouts/_case_detail_page.html.slim
misaka: Ok, done!

app/views/layouts/_case_detail_page.html.slim
misaka: Added!

PR #106, opened 2017-02-24
CT-474 assigners only see final responses

misaka:
https://dsdmoj.atlassian.net/browse/CT-474

*Ignore commits for CT-416, this branch is based off of the CT-416 branch which isn't, as of this time, a part of develop*

CT-474 specs out that assigners only see the uploaded files list after the case
has been marked as responded-to, but really what this means is hiding the
uploaded files list for the assigners: they'll see the uploaded files in the
Response Details section which will be added as part of CT-484 et al.

This commit includes some minor refactoring of specs like creating shared
examples and bubbling up fixtures that are used in multiple rspec contexts.


spec/features/case_uploaded_files_spec.rb
```diff
@@ -2,72 +2,96 @@
...
 
-        expect {
-          case_details_page.uploaded_files.files.first.download.click
-        }.to redirect_to_external(attachment_url)
-      end
+    scenario 'can remove the response' do
+      case_details_page.load(id: kase.id)
+
+      case_details_page.uploaded_files.files.first.remove.click
+      expect(case_details_page).not_to have_uploaded_files
```
aliuk2012: You could also have this
```expect(case_details_page).to have_no_uploaded_files```

PR #109, opened 2017-02-28
Refactor feature specs organise by resource

misaka:
Here's the structure that @aliuk2012 and I worked out for the feature specs in this project:

```
features
  | case 
    | foi
      | accept a case
      | assign a case
      | case details
        | sidebar
        | description 
        | uploads
        | response
      | close a case 
      | create a case 
      | listing cases - workbasket
      | reject a case 
      | respond to a case 
    | gq
  | common
    | sign in 
    | sign out
    | feedback
```

Using the above as a guide, I've renamed the feature specs as such:

```
spec/features/03_assignment/assign_case_from_detail_view.rb               -> spec/features/cases/foi/case_assigning_spec.rb
spec/features/case_deadlines_spec.rb                                      -> spec/features/cases/foi/case_viewing/deadlines_spec.rb
spec/features/case_uploaded_files_spec.rb                                 -> spec/features/cases/foi/case_viewing/uploaded_files_spec.rb
spec/features/case_detail_view_spec.rb                                    -> spec/features/cases/foi/case_viewing/viewing_spec.rb
spec/features/05_close/close_case_spec.rb                                 -> spec/features/cases/foi/case_closing_spec.rb
spec/features/02_create/permissions_spec.rb                               -> spec/features/cases/foi/case_creating/permissions_spec.rb
spec/features/02_create/case_create_spec.rb                               -> spec/features/cases/foi/case_creating/creating_spec.rb
spec/features/case_list_view_spec.rb                                      -> spec/features/cases/foi/case_isting_spec.rb
spec/features/04_respond_to_correspondence/mark_response_as_sent_spec.rb  -> spec/features/cases/foi/case_responding/mark_as_sent_spec.rb
spec/features/04_respond_to_correspondence/upload_response_spec.rb        -> spec/features/cases/foi/case_responding/upload_response_spec.rb
spec/features/03_assignment/respond_to_assignment_spec.rb                 -> spec/features/cases/foi/case_assignment_responding_spec.rb
spec/features/00_feedback/submit_feedback_spec.rb                         -> spec/features/common/feedback_spec.rb
spec/features/01_sign_in/sign_in_spec.rb                                  -> spec/features/common/sign_in_spec.rb
spec/features/01_sign_in/sign_out_spec.rb                                 -> spec/features/common/sign_out_spec.rb
```

In addition, I've also refactored the SitePrism page objects, trying to keep them in sync with what we have in `correspondence_tool_public`:

```
spec/support/pages/case_details_page.rb  -> spec/site_prism/page_objects/pages/cases/show_page.rb
spec/support/pages/case_list_page.rb     -> spec/site_prism/page_objects/pages/cases_page.rb
spec/support/pages/case_response_page.rb -> spec/site_prism/page_objects/pages/cases/respond_page.rb
spec/support/pages/case_upload_page.rb   -> spec/site_prism/page_objects/pages/cases/new_response_upload_page.rb
spec/support/pages/feedback_section.rb   -> spec/site_prism/page_objects/sections/service_feedback_section.rb
spec/support/pages/login_page.rb         -> spec/site_prism/page_objects/pages/login_page.rb
```


spec/rails_helper.rb
```diff
@@ -20,6 +20,9 @@
 # end with _spec.rb. You can configure this pattern with the --pattern
 # option on the command line or in ~/.rspec, .rspec or `.rspec-local`.
 Dir[Rails.root.join("spec/support/**/*.rb")].each { |f| require f }
+$LOAD_PATH.unshift(File.join(File.expand_path('..', __FILE__)))
+require 'page_objects/all_pages.rb'
```
aliuk2012: Why are you requiring this?

aliuk2012:
could you change the directory name from page_objects to site_prism? So all the pages/sections should be placed under spec/site_prism not spec/page_objects?
The module should be AnswerTool not PageObjects

misaka:
@aliuk2012 The reason I didn't use `site_prism` is that then the pages would be put into the namespace `SitePrism`. We could use `site_prism/page_objects` and then just use `PageObjects` as the top-most namespace name?

spec/rails_helper.rb
misaka: It won't come in otherwise I don't think, have you tested this and found otherwise?

aliuk2012: in the ask tools rails_helper.rb  all we have is 
```
# Load Site Prism
Dir[Rails.root.join('spec/site_prism/**/*.rb')].each { |f| require f }
```

misaka: Right, but I ran into a problem where `pages` got included before `sections`, causing errors since `pages` try to use `sections` which is why on line 25 below we order the includes to pull in `sections` before `pages`.

misaka: Just realised I didn't fully explaining this. Because I had to order `sections` before `pages`, I modified the glob from `spec/page_objects/**/*.rb` to be `spec/page_objects/{sections,pages}/**/*.rb`. But that means that `all_pages.rb` isn't required since `sections` and `pages` are being picked out. I did try `spec/page_objects/{all_pages.rb,sections,pages}/**/*.rb` but ran into problems, probably because of the `/**/*.rb` bit, so instead I just pulled it out on it's own, which is what you see here.

aliuk2012: Ah, I see. You right the thing with Ask tool is that ATM this is no sections. Could we maybe just copy/paste/inline the sections for the time being? so all we are dealing with are pages

misaka: I agree that those that can be inlined, should be. But some of the `sections` are being re-used ... and in fact there are some `sections` that are inline that should be pulled out (like, for example, the `uploaded_files` on the `case_details_page`, which should not show up on certain pages).

misaka:
After speaking with @aliuk2012 I'm going to do one more minor renaming to convert the names to be more action-based ... e.g. `case_closing_spec.rb`.

PR #110, opened 2017-03-02
CT-472 Drafters  accept/reject journey

aliuk2012:
https://dsdmoj.atlassian.net/browse/CT-472

- Adds a helper method to generate the case number link 
- Removes the "Accept/Reject" button from the case detail page
- Redirects to the Accept/Reject page is a drafter manually edits the url to try access the case detail page.


app/controllers/cases_controller.rb
```diff
@@ -36,6 +36,9 @@ def create
 
   def show
     set_permitted_events
+    if policy(@case).can_accept_or_reject_case?
+      redirect_to edit_case_assignment_path @case, @case.assignments.last.id
+    end
```
misaka: If I'm not mistaken, the line below will get run in the case of a redirect since `redirect_to` doesn't actually exit the show method. I know it's inconsequential, but I think this would be better expressed with an if/else with the line below being in an `else` block.

misaka: If one day we have more than one kind of case assignments, for example being assigned to a `case_worker`, then I suspect we risk having a subtle bug here where someone could get redirected to the wrong assignment, say if two assignments are created with the `case_worker` assignment coming after the `drafter` one. Tests would have to be specifically designed to test for this, which I'm worried we would forget to do. Is there a cheap way for us to modify line 40 so that it doesn't just take the last assignment, but specifically looks for one that is appropriate for the user that is viewing the page?

Just to be clear, I don't think we absolutely have to protect for this, if say for example it's going to take a lot of work, I'm just hoping there's a relatively cheap way to do it.

misaka: Where are we testing this redirect, btw? I didn't see it in the specs below, did I miss it?

app/helpers/cases_helper.rb
```diff
@@ -28,4 +24,14 @@ def action_button_for(event)
           class: 'button', method: :patch
     end
   end
+
+  def case_detail_link(kase)
+    if kase.current_state == "awaiting_responder" && @user.drafter?
```
misaka: This only checks if the case is awaiting a responder and the user is a drafter? Shouldn't it require that the drafter is the one that has been assigned to the case? Or is that sorted out later?

spec/features/case_detail_view_spec.rb
```diff
@@ -122,20 +122,5 @@
...
-    end
-
-    scenario 'the assigned drafter views the case' do
-      login_as foi.drafter
-      page.load id: foi.id
-
-      expect(page.sidebar.actions).
-        to have_link('Accept or reject',
-          href: edit_case_assignment_path(foi, foi.assignments.last))
-    end
```
misaka: Seems to me we should be testing that the redirect happens here instead.

app/helpers/cases_helper.rb
aliuk2012: We dont care who the drafter is or if they are the one assigned to the case. Remember this will be in the case list which we have already limited drafters case lists to case assigned to them. I suppsect this will change quiet a bit once we start implementing business units and letting drafters see all the cases assigned to their business units. We also have this policy in place https://github.com/ministryofjustice/correspondence_tool_staff/blob/develop/app/policies/case_policy.rb#L11.

aliuk2012:
After speaking to @misaka, I decided to change the approach I took on completing this piece of work.

I stripped out the helper method because it didn't seem to be fit for purpose in that it was merely returning a different url based on some conditions. Instead it would have been a far better approach to implement it at the Controller/Action level. The Action would decided whether to redirect the user or render the page based on various conditions.

As part of the new implementation, I decided to create a new policy to risk access to the case detail page which would stop users editing url to can access to case details that they were not supposed to see. This policy is checked before it checks to see if the page should be rendered or whether it should redirect a user.

spec/features/case_detail_view_spec.rb
stephenrichards: Let's get this merged and raise a ticket to put a test in for this.

misaka: Tests appear to be in the controller specs so that works for me.

PR #112, opened 2017-03-06
CT-483 create close case page

stephenrichards:
This PR is the first stage in the Close Case story.

It creates a new STI table CaseClosureMetadata which holds three types of data: outcomes, refusal reasons, and exemptions, all of which are used in the close page.

This PR only takes it as far as recording the outcome.  Refusal reasons and Exemptions to follow.


app/views/cases/close.html.slim
```diff
@@ -0,0 +1,25 @@
...
+
+  = form_for @case, url: process_closure_case_path do |f|
+    .form-group
+      = f.gov_uk_date_field :date_responded, { legend_text: t('.close_date'), form_hint_text: t('.date_example') }
+      br
+      = f.radio_button_fieldset :outcome,
+              choices: CaseClosure::Outcome.all.map(&:name)
+
+    = f.submit t('button.continue'), {class: 'button'}
+
```
aliuk2012: Could you replace all the mark up in this file so the file looks like this?

```
.grid-row
  section.case-main
    h1.case-heading
      .case-heading--secondary
        span.visually-hidden
          = t('common.case.header_case_number')
        = @case.number
        = " | "
        = @case.subject
      = t('.page_heading')

= form_for @case, url: process_closure_case_path do |f|
  .form-group
    = f.gov_uk_date_field :date_responded, { legend_text: t('.close_date'), form_hint_text: t('.date_example') }

  = f.radio_button_fieldset :outcome, choices: CaseClosure::Outcome.all.map(&:name)

  = f.submit t('button.continue'), {class: 'button'}
```

app/models/case.rb
```diff
@@ -148,8 +150,11 @@ def respond(current_user_id)
     drafter_assignment.destroy
   end
 
-  def close(current_user_id)
-    state_machine.close!(current_user_id)
+  def close(current_user_id, case_closure_params)
+    self.transaction do
```
misaka: :100:

app/views/cases/close.html.slim
stephenrichards: Done.

aliuk2012:
Not huge problem but I was a bit surprised to see the migration and all the refusal/exception data added in this pr because its not played back on the view or used anywhere else in the close case page at this point in time. I would have expected to see if as part of the Refusal reasons and Exemptions work which is going to be added later. Its harder to check and test that the data etc is working and correct.

It would have helped to keep the size of the pr down and would have been easier to review.

spec/features/cases/foi/case_closing_spec.rb
```diff
@@ -17,6 +25,11 @@
       to have_link('Close case', href: close_case_path(kase))
     click_link 'Close case'
 
+    expect(cases_close_page).to have_content("Close case")
+    cases_close_page.fill_in_date_responded(2.days.ago)
```
misaka: I don't think we test that this date_responded actually makes it into the DB, neither here nor in the controller test.

spec/site_prism/page_objects/pages/cases/close_page.rb
```diff
@@ -0,0 +1,27 @@
+module PageObjects
+  module Pages
+    module Cases
+      class ClosePage < SitePrism::Page
+        set_url '/cases/{id}'
+        # TODO: the sections here ... actually this whole page, can show up in other
```
misaka: Looks like an old comment left-in from `show_page.rb` that needs to be removed.

spec/features/cases/foi/case_closing_spec.rb
stephenrichards: have added it to the controller spec

spec/site_prism/page_objects/pages/cases/close_page.rb
stephenrichards: Done

PR #115, opened 2017-03-07
CT-484 add response details section

misaka:
https://dsdmoj.atlassian.net/browse/CT-484

Add the response details section to the case details page as per the design. This includes a section for uploaded files and for the responder's name.

The commits `fix user metadata in responded case transition` and `Generate fullname from email when adding fullname to User` come from work that's been broken out into other PRs, so ignore them if you see them here.


app/policies/case_policy.rb
```diff
@@ -23,6 +23,13 @@ def can_accept_or_reject_case?
     self.case.awaiting_responder? && self.case.drafter == user
   end
 
+  def can_remove_attachment?
+    case self.case.current_state
+    when 'drafting' then self.case.drafter == user
+    when 'responded' then user.assigner?
```
aliuk2012: Do we know whether assigners should be able to remove uploaded responses?

stephenrichards: The answer is: NO

stephenrichards: IN fact, NO-ONE should be able to remove an uploaded file once the case has been marked as reponded

misaka:
Just to comment on the specs, in case it's not obvious I've done a couple changes to them:

- broken out the `uploaded_files_spec` into three specs:
  - `viewing_response_details_spec`
  - `downloading_response_spec`
  - `removing_response_spec`
- reorganised the structure of many of the specs in these files so that the scenario isn't too honking huge and to try to remove a little bit of the duplication ... not because tests need to be totally DRY (they don't) ... but to try to reduce some of the noise in the spec features. This is the structure I used:
  - context of logged in as assigner/drafter
    - context of case state (drafting, responded, etc)
      - scenario

app/policies/case_policy.rb
```diff
@@ -23,6 +23,13 @@ def can_accept_or_reject_case?
     self.case.awaiting_responder? && self.case.drafter == user
   end
 
+  def can_remove_attachment?
```
aliuk2012: We need an "can_remove_attachment" translation in the locales file.

misaka: Added.

PR #122, opened 2017-03-13
Add 'cts' script.

misaka:
Implement project-specific actions as a stand-alone cli.

Normally we'd add stuff like this to a rake task, but there are a few reasons why Rake isn't the best tool for the job: rake -T is slow and passing command-line options is painful are the most obvious ones. So taking a page out of Rails' book, here is a Thor-based cli where we can put project-specific commands.


misaka:
@stephenrichards has added a different `cts` script, I'll look to merge what I've got here into that one at some point. Maybe.

PR #123, opened 2017-03-15
CT-485 add reason for refusal to response details

misaka:
https://dsdmoj.atlassian.net/browse/CT-485

Finishes up some of the work for CT-485 to add reason for refusal to response details. Also added some timeline-related info to the response details as part of CT-484


spec/site_prism/page_objects/pages/cases/show_page.rb
```diff
@@ -37,7 +37,10 @@ class ShowPage < SitePrism::Page
           end
           element :responder, '#request--responder'
           element :date_responded, '#request--date-responded'
+          element :timeliness, '#request--response-timeliness'
+          element :time_taken, '#request--response-time_taken'
```
aliuk2012: only minor, why underscore in the id when others are hypenated

misaka: Whoops!

misaka: Fixed. Changed the history to keep it clean.

PR #125, opened 2017-03-15
CT-510 Add "no files" uploaded validation

aliuk2012:
https://dsdmoj.atlassian.net/browse/CT-510

On the new response upload page, if you clicked "Upload response"
without selecting any files to upload. It would throw an error message
that was not caught or dealt with.

This solution takes the user back to the page and presents them
with a flash message telling them to select some files to upload.


spec/controllers/cases_controller_spec.rb
```diff
@@ -842,6 +842,19 @@ def do_upload_responses
...
+
+      context 'does not upload any valid files' do
+
+        it 'Returns error message if not files uploaded' do
+          post :upload_responses, params: {
+              id:             kase,
+              type:           'response'
+          }
+          expect(response).to have_rendered(:new_response_upload)
+
```
misaka: Shouldn't we check for the flash message here too?

aliuk2012: 👍  Good point

PR #128, opened 2017-03-21
CT-562 Only run DB clear and reseed rake tasks if in a safe environment

stephenrichards:
Provide rake HostEnv.safe method that will only run the block if rails env is development or test, or the hosting environment is staging or dev, i.e. it cannot be run on the live environemtn.

The db:clear (used by db:reseed) and cases:demo_entries changed to run inside the HostEnv.safe method.

Also fixed a bug which prevented tests from working on Mondays or Tuesdays ;~)


spec/features/cases/foi/case_closing_spec.rb
```diff
@@ -76,7 +76,8 @@
       click_link 'Close case'
 
       expect(cases_close_page).to have_content("Close case")
-      cases_close_page.fill_in_date_responded(2.days.ago)
+      
+      cases_close_page.fill_in_date_responded(2.business_days.ago)
```
aliuk2012: Why are these tests being changed as part of this PR?

spec/features/cases/foi/case_closing_spec.rb
```diff
@@ -87,7 +88,7 @@
       expect(cases_show_page.sidebar.actions.text).to eq 'No actions available'
 
       expect(cases_show_page.response_details.date_responded.text)
-        .to eq 2.days.ago.strftime('%e %b %Y').strip
+        .to eq 2.business_days.ago.strftime('%e %b %Y').strip
```
aliuk2012: Why are these tests being changed as part of this PR?

spec/rake_task_helpers/host_env_spec.rb
```diff
@@ -0,0 +1,208 @@
+require 'rails_helper'
+require File.join(Rails.root, 'lib', 'rake_task_helpers', 'host_env')
+
+describe HostEnv do
+
+  context 'development rails environment' do
+    before(:each) do
+      ENV['RAILS_ENV'] = 'development'
```
misaka: Not 100% it's clearer, but you can get rid of the `after` block if you mock out `ENV`:

```ruby
allow(ENV).to receive(:[]).with('RAILS_ENV').and_return('development')
```

spec/features/cases/foi/case_closing_spec.rb
aliuk2012: Sorry I hadn't seen this part:

> Also fixed a bug which prevented tests from working on Mondays or Tuesdays ;~)

I think that should have been a separate commit. I'm not sure if you want to maybe remove this from your PR as I'm working on the closing case feature specs and have added these two fixes. If not then there might be some merge conflicts that I'll have to sort out.

spec/features/cases/foi/case_closing_spec.rb
aliuk2012: Sorry I hadn't seen this part:

> Also fixed a bug which prevented tests from working on Mondays or Tuesdays ;~)

I think that should have been a separate commit. I'm not sure if you want to maybe remove this from your PR as I'm working on the closing case feature specs and have added these two fixes. If not then there might be some merge conflicts that I'll have to sort out.

PR #138, opened 2017-03-28
CT-557 Additional state - awaiting_dispatch to the case

stephenrichards:
that is triggered once the first response is uploaded.  COnversley,
if the last response is deleted, the state reverts back to drafting.

Cases can only be marked as responsed from awaiting dispatch state.


lib/case_state_machine.rb
```diff
@@ -40,6 +46,13 @@ def add_responses!(assignee_id, filenames)
              event:       :add_responses
   end
 
+  def revert_to_drafting!
```
misaka: There are no tests for this in `case_state_machine_spec.rb`, we should have some.

lib/case_state_machine.rb
```diff
@@ -21,11 +22,16 @@ class CaseStateMachine
   end
 
   event :add_responses do
-    transition from: :drafting, to: :drafting
+    transition from: :drafting, to: :awaiting_dispatch
+    transition from: :awaiting_dispatch, to: :awaiting_dispatch
+  end
+
+  event :revert_to_drafting do
```
stephenrichards: OK, done.  Couldn't get guard transition to work, but achived the same result by having two separate events, remove attachment and remove_last_attachement.

lib/case_state_machine.rb
```diff
@@ -40,6 +46,13 @@ def add_responses!(assignee_id, filenames)
              event:       :add_responses
   end
 
+  def revert_to_drafting!
+    trigger! :revert_to_drafting,
+             event: :revert_to_drafting
```
stephenrichards: done

app/policies/case_policy.rb
```diff
@@ -8,7 +8,7 @@ def initialize(user, kase)
   end
 
   def can_add_attachment?
-    self.case.drafting? && self.case.drafter == user
+    (self.case.drafting? || self.case.awaiting_dispatch?) && self.case.drafter == user
```
stephenrichards: yep.  done

app/policies/case_policy.rb
```diff
@@ -25,13 +25,13 @@ def can_accept_or_reject_case?
 
   def can_remove_attachment?
     case self.case.current_state
-    when 'drafting' then self.case.drafter == user
+    when 'drafting', 'awaiting_dispatch' then self.case.drafter == user
```
stephenrichards: Actually this is wrong.  only cases in awaiting_dispatch can have attachments remove.

lib/case_state_machine.rb
```diff
@@ -46,13 +50,21 @@ def add_responses!(assignee_id, filenames)
              event:       :add_responses
   end
 
-  def revert_to_drafting!
-    trigger! :revert_to_drafting,
-             event: :revert_to_drafting
+  def remove_response(assignee_id, filename, num_responses)
```
misaka: This method and the following one has separated `add_responses!` from `add_responses` below, can we keep them together please?

PR #139, opened 2017-03-28
WIP: CT-353 change case assignment to use radio buttons

misaka:
https://dsdmoj.atlassian.net/browse/CT-353

This is a temporary change until we sort out how to do the KILO map, previously
it was a select list which was only done because of technical problems with the
GOVKUK form builder.

Also, this change relies on ministryofjustice/govuk_elements_form_builder#83 and there is a chance the way we build the radio buttons will change, as per the comments in that PR.


misaka:
Replacing with #144

PR #141, opened 2017-03-29
CT-544 change case policies to accommodate teams

misaka:
https://dsdmoj.atlassian.net/browse/CT-544

Continuation of the work to convert the system to use teams. Policies needed to be changed variously:

- check that the logged-in user's is part of the responding team assigned to the case
- rename drafter to responder and assigner to manager

And, plus, a gazillion tests had to change.


misaka:
Replacing with #144

PR #148, opened 2017-04-07
CT-572 Convert attachments to PDF so they can be previewed

stephenrichards:
This PR is a combination of the following tickets:

- CT-572 Background job to convert attachments to PDF
- CT-573 Trigger the background job after adding an attachment


app/models/case_attachment.rb
```diff
@@ -35,8 +36,54 @@ def temporary_url
...
+    preview_file.close
+    preview_file.path
+  end
+
+
+  def upload_preview(filepath)
+    pdf_key = "#{self.case.attachments_dir('response_previews')}/#{File.basename(key, File.extname(key))}.pdf"
+    preview_object = CASE_UPLOADS_S3_BUCKET.object(pdf_key)
+    result = preview_object.upload_file(filepath)
+    result == true ? pdf_key : nil
```
misaka: The return value isn't being used. Not that we shouldn't return a value, but if `result == false` would it be worth it to raise?

app/models/case_attachment.rb
```diff
@@ -35,8 +36,54 @@ def temporary_url
...
+    else
+      original_filepath = download_original_file
+      preview_filepath = make_preview_filename
+      begin
+        Libreconv.convert original_filepath, preview_filepath
+        self.preview_key = upload_preview(preview_filepath)
+      rescue StandardError => err
+        Rails.logger.error "Error converting CaseAttachment #{self.id} to PDF"
+        Rails.logger.error "#{err.class} - #{err.message}"
+        Rails.logger.error err.backtrace
```
misaka: Why not just re-raise `err` here? Don't we risk not doing the right thing by trying to output the message and backtrace ourself, instead of letting the default handler go it?

spec/jobs/pdf_maker_job_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    errmsg = 'PdfMakerJob Error creating Preview for attachment 123: ActiveRecord::RecordNotFound - ActiveRecord::RecordNotFound'
+    expect(Rails.logger).to receive(:error).with(errmsg)
+
+    described_class.perform_now(123)
+  end
+
+  after do
+    clear_enqueued_jobs
+    clear_performed_jobs
+  end
```
misaka: We should test the unhappy case here where there was an exception raised while making the preview, where `preview_key` gets set to `ERROR` or `nil` or whatever is appropriate.

spec/models/case_attachment_spec.rb
```diff
@@ -93,4 +94,86 @@
...
+        # let(:preview_key) { '' }
+        let(:s3_object) { double 'S3 Object' }
+        let(:kase) { jpg_case_attachment.case }
+
+        it 'uploads the file to s3' do
+          expect(CASE_UPLOADS_S3_BUCKET).to receive(:object).with("#{kase.id}/response_previews/my_photo.pdf").and_return(s3_object)
+          expect(s3_object).to receive(:upload_file).with('xxx')
+
+          jpg_case_attachment.__send__(:upload_preview, 'xxx')
+        end
```
misaka: Similar to the comment on tests for `PdfMakerJob#perform`, we should test that `upload_preview` returns an error in the unhappy case. Also, we should bring those two methods in line, one returns `ERROR` and the other returns `nil`.

app/models/case_attachment.rb
stephenrichards: Exceptions will get caught in the parent block.  Have  changed the code so that if upload_file returns false, it will queue another job, with the number of times it's tried, and there is a stop to prevent it being retried more than 5 times.

spec/jobs/pdf_maker_job_spec.rb
stephenrichards: done

spec/models/case_attachment_spec.rb
stephenrichards: done

PR #161, opened 2017-04-20
CT-633 Show rejection reasons

stephenrichards:
This is implemented by showing a case history table on all detail views


spec/controllers/cases_controller_spec.rb
```diff
@@ -327,6 +330,11 @@ def case_closure_params(kase)
         it 'renders the show template' do
           expect(response).to render_template(:show)
         end
+
+        it 'decorates the collection of case transitions' do
+          expect(assigns(:case_transitions)).to be_an_instance_of(Draper::CollectionDecorator)
+          expect(assigns(:case_transitions).map(&:class)).to eq [ CaseTransitionDecorator ]
+        end
```
misaka: Can we find a more sensible place to put this? This isn't really about the manager's view directly, is it.

stephenrichards: Well, it's about any authorised user's view.  Where would you suggest?

spec/controllers/cases_controller_spec.rb
```diff
@@ -36,6 +36,10 @@
         get :edit, params: { id: first_case }
         expect(response).to redirect_to(new_user_session_path)
       end
+
+      it 'decorates the case transitions collection' do
```
stephenrichards: Have removed it from here.  See below.

PR #162, opened 2017-04-20
CT-422 Add primary navigation

aliuk2012:
https://dsdmoj.atlassian.net/browse/CT-422


app/views/layouts/_global_nav.html.slim
```diff
@@ -0,0 +1,4 @@
+nav.global-nav
+  ul
+    li class="global-nav-item"
+      = link_to t('nav.cases'), cases_path, {class:"#{'active' if current_page?(cases_url ) || current_page?(root_url)}"}
```
misaka: Why put the class inside of a string here? I think a ternary would be clearer:

```
class: current_page?(cases_url ) || current_page?(root_url) ? 'active' : ''
```

app/views/layouts/application.html.slim
```diff
@@ -29,6 +29,9 @@
 = content_for :content do
   = csrf_meta_tags
   = render partial: 'layouts/phase_banner'
+  - if controller_name != 'sessions' && @user.full_name.present?
```
misaka: Why `@user.full_name.present?`?

PR #163, opened 2017-04-21
CT-531 add approver role to models and db

misaka:
https://dsdmoj.atlassian.net/browse/CT-531

Supporting work required for the epic to add clearance tests.


aliuk2012:
I think we are missing Factories for the new roles/teams and I think we should also include seeding methods for these new role/team

misaka:
@aliuk2012 -- @stephenrichards says that he'll add the factories ... makes sense, he'll be the first one to need them presumably.

PR #167, opened 2017-04-24
CT-323 Refactor state machine so that state is held on case record

stephenrichards:
This PR is preparation work for the real work involved in ticket CT-323 which
is to show Closed cases for an assigner or drafter.

This is necessary so that we can easily select cases in a particular state (in this
case, closed).


app/models/case.rb
```diff
@@ -246,6 +247,13 @@ def has_ncnd_exemption?
 
   private
 
+  def set_initial_state
+    if self.current_state.nil?
```
misaka: Forgot to take this out.

PR #168, opened 2017-04-25
CT-458 Add Today button to new case page

stephenrichards:
This PR adds a Today button, which when pressed, fills in the date field boxes with today's date.


app/views/cases/new.html.slim
```diff
@@ -6,27 +6,27 @@
...
+    = f.text_area :postal_address, {rows: 4}
+
+    = f.radio_button_fieldset :requester_type,
+                                choices: [:academic_business_charity,
+                                          :journalist,
+                                          :member_of_the_public,
+                                          :offender,
+                                          :solicitor,
+                                          :staff_judiciary,
+                                          :what_do_they_know]
```
aliuk2012: Why are there so many changes to this file? I would have only expected the govuk_date_fields lines to be updated

PR #169, opened 2017-04-25
CT-649 adding state transitions for approvers

misaka:
https://dsdmoj.atlassian.net/browse/CT-649


app/controllers/cases_controller.rb
```diff
@@ -119,18 +119,7 @@ def search
...
-        policy(@case).can_assign_case?
-      when :add_responses
-        policy(@case).can_add_attachment?
-      when :respond
-        policy(@case).can_respond?
-      when :close
-        policy(@case).can_close_case?
-      end
-    end
+    @permitted_events = @case.state_machine.permitted_events(current_user.id)
```
stephenrichards: I Like this moved out of the controller

spec/lib/case_state_machine_spec.rb
```diff
@@ -1,18 +1,96 @@
 require 'rails_helper'
 
+RSpec::Matchers.define :trigger_the_event do |event|
```
stephenrichards: let's put these matchers in spec/support/matchers

PR #170, opened 2017-04-25
CT-323 Closed Case Page

stephenrichards:


app/views/cases/closed_cases.html.slim
```diff
@@ -0,0 +1,30 @@
...
+              = " "
+              = link_to kase.number, case_path(kase.id)
+            td aria-label="#{t('.name-subject')}"
+              <strong>
+              = kase.name
+              </strong>
+              <br>
+              <span>
+              = kase.subject
+              </span>
```
aliuk2012: Lines 24 to 30 should be in slim not HTML

config/locales/en.yml
```diff
@@ -208,6 +208,11 @@ en:
       case_created: Case successfully created
 
   cases:
+    closed_cases:
+      page_heading: Closed Cases
```
aliuk2012: Lowercase 'c' for cases

config/locales/en.yml
```diff
@@ -258,6 +263,7 @@ en:
 
   nav:
     cases: Cases
+    closed_cases: Closed Cases
```
aliuk2012: Lowercase 'c'

config/locales/en.yml
```diff
@@ -258,6 +263,7 @@ en:
 
   nav:
     cases: Cases
```
aliuk2012: Shouldn't this change to "Open cases"?

app/views/cases/closed_cases.html.slim
```diff
@@ -0,0 +1,30 @@
...
+        - @cases.each do |kase|
+          tr.case_row
+            td aria-label="#{t('.number')}"
+              span.visually-hidden
+                = t('.view_case')
+              = " "
+              = link_to kase.number, case_path(kase.id)
+            td aria-label="#{t('.name-subject')}"
+              <strong>
+              = kase.name
```
aliuk2012: should be tabbed

app/views/cases/closed_cases.html.slim
```diff
@@ -0,0 +1,30 @@
...
+              </strong>
+              <br>
+              <span>
+              = kase.subject
```
aliuk2012: should be tabbed

config/locales/en.yml
stephenrichards: correct

app/views/cases/closed_cases.html.slim
stephenrichards: correct

config/locales/en.yml
stephenrichards: correct

config/locales/en.yml
stephenrichards: yep

app/views/cases/closed_cases.html.slim
stephenrichards: Addressed by converting to slim

app/views/cases/closed_cases.html.slim
stephenrichards: Addressed by converting to slim

stephenrichards:
I've written a view spec for the Global Nav Partial.  I think this, along with GlobalNavManager spec means that we can do away with most of the Primary Navigation feature spec, retaining the bits that check that it's not visible at all on the login page.

In the Global Nav spec we test that the correct nav links are set up depending on role (or will do in the future), and in the view spec, we test that the right links are produced given a particular GlobalNavManager set up, and that the link is marked as active if the current_page?(url) returns true.

That should be enough, I think; we're testing the mechanism, and therefore don't have to test each and every possible combination.

Let me know what you think.

misaka:
@stephenrichards Agreed that we can start to replace bloated feature tests that test every single and detail of the views with view specs, and focus primarily on user journeys in the feature tests.

PR #171, opened 2017-04-27
CT-625 case view for disclosure team

misaka:
https://dsdmoj.atlassian.net/browse/CT-625

Provide a case list view specific for approvers. This includes adding an Incoming cases tab  (and the machinery to configure these per-user), and some testing framework that wasn't there before.

This is currently only available to approvers, as configured in the `settings.yml` file.  Also, the Incoming cases tab is identical to the Open cases tab because the Open cases isn't filtering down to only show those cases that are assigned to a user in the team, that will have to be another ticket.


app/controllers/cases_controller.rb
```diff
@@ -15,6 +15,12 @@ def closed_cases
     @cases = policy_scope(Case.closed).map(&:decorate)
   end
 
+  def incoming_cases
+    team_cases = Case.with_team(*current_user.teams)
+                   .without_user_for_team(*current_user.teams)
+    @cases = policy_scope(team_cases).map(&:decorate)
```
stephenrichards: I don't think the without_user_for_team scope is required here because the with_team scope only returns non-accepted/non-rejected claims.  Perhaps a better name for the scope would be unaccepted_for_team, or something along those lines

app/models/user.rb
```diff
@@ -34,27 +34,35 @@ class User < ApplicationRecord
...
+    joins(:team_roles).where(teams_users_roles: { role: 'approver' })
   }
 
-  # def manager?
-  #   team_roles.managing_teams.include? self
-  # end
-
   def manager?
     not managing_teams.empty?
+    # team_roles.managing_teams.include? self
```
stephenrichards: managing_teams.any?

app/models/user.rb
```diff
@@ -34,27 +34,35 @@ class User < ApplicationRecord
...
   end
 
   def responder?
     not responding_teams.empty?
   end
+
+  def approver?
+    not approving_teams.empty?
+  end
```
stephenrichards: prefer to see any? rather than not empty

spec/models/user_spec.rb
```diff
@@ -28,4 +28,56 @@
...
+  let(:responder) { create :responder }
+  let(:approver)  { create :approver }
+
+  describe '#manager?' do
+    it 'returns true for a manager' do
+      expect(manager.manager?).to be true
+    end
+
+    it 'returns true for a responder' do
+      expect(responder.manager?).to be false
```
stephenrichards: it 'returns false'

spec/models/user_spec.rb
```diff
@@ -28,4 +28,56 @@
...
+    end
+
+    it 'returns true for an approver' do
+      expect(approver.manager?).to be false
```
stephenrichards: false

spec/models/user_spec.rb
```diff
@@ -28,4 +28,56 @@
...
+    it 'returns true for a manager' do
+      expect(manager.responder?).to be false
+    end
+
+    it 'returns true for a responder' do
+      expect(responder.responder?).to be true
+    end
+
+    it 'returns true for an approver' do
+      expect(approver.responder?).to be false
```
stephenrichards: false - and below...

PR #174, opened 2017-05-04
CT-642 needs clearance

misaka:
https://dsdmoj.atlassian.net/browse/CT-642

When DACU BMT creates a case they can flag it as requiring clearance from DACU a Disclosure Specialist. This change implements that on the case create page, and includes some tweaks for existing-but-never-used code in the back-end for these "trigger" cases.


misaka:
@aliuk2012 When the "flag for clearance" hasn't been selected by the user, I've made the error look like this:

> Does a disclosure specialist need to see this? Please choose yes or no.

Based on the existing wording "Does a disclosure specialist need to see this?" I just added the prompt "Please choose yes or no." ... is that good enough?

aliuk2012:
I think that message is fine, I believe Kellie and Martin want to do an error message copy review so if its an issue it will be sorted out then

app/controllers/cases_controller.rb
```diff
@@ -26,7 +26,14 @@ def create
     authorize Case, :can_add_case?
 
     @case = Case.new(create_foi_params)
-    if @case.save
+    if !create_foi_params[:flag_for_disclosure_specialists].present?
+      @case.valid?
```
stephenrichards: Isn't this better expressed as if create_foi_params[:flag_for_disclosure_specialists].blank?

app/controllers/cases_controller.rb
```diff
@@ -26,7 +26,14 @@ def create
...
+      @case.errors.add(:flag_for_disclosure_specialists, :blank)
```
stephenrichards: Why are we doing this in the controller?  Surely this is better in the validation routines of the correspondence model?

app/controllers/cases_controller.rb
misaka: Yup, will tweak!

app/controllers/cases_controller.rb
misaka: It doesn't feel like the flagging for clearance belongs in the model. There is no actual `flag_for_clearance` attribute on the model, we flag by assigning approvers to the case, and when we do flag for clearance we need the `current_user` which isn't, and probably shouldn't, be available in the model anyway (see line 35 below).

I've gone back and forth a couple times on this, but when weighing up exposing `current_user` in the model versus adding to the errors here, I decided this was cleaner.

misaka: Sorry, I should've said in the first sentence that it doesn't feel like handling the `flag_for_clearance` attribute belongs in the model.

PR #175, opened 2017-05-05
CT-653 Sort case history

aliuk2012:
https://dsdmoj.atlassian.net/browse/CT-653

- Sorts the case history so the last entry is first
- Updates the govuk date fields gem to use the latest one to fix issues with "Today" button


misaka:
@aliuk2012 I was having a look at the Statesman docs because I know they provide a `sort` field, or something like that, and noticed that they actually have a [`history` method](https://github.com/gocardless/statesman#machinehistory) intended to do exactly what we're doing here.

PR #178, opened 2017-05-09
CTS script to seed database with cases in various states

stephenrichards:


db/seeders/case_seeder.rb
```diff
@@ -0,0 +1,180 @@
...
+
+    args.each { |arg| process_arg(arg) }
+
+    if @invalid_params
+      puts "Program terminating"
+      exit 1
+    end
+  end
+
+  def process_arg(arg)
```
misaka: Thor provides a way to specify cmdline options, wouldn't it be cleaner to use that?

https://github.com/erikhuda/thor/wiki/Method-Options

PR #180, opened 2017-05-11
CT-646 Create mechanism for Disclosure team to view flagged orders

stephenrichards:
This PR implements the Db scoping necessary to produce the incoming and open pages for 
DACU disclosure team.

The index, closed_cases and incoming_cases actions on the cases_controller now instantiate a 
CaseFinderService object which has the responsibilty of returning the correct set of cases in the correct order.


misaka:
This has been merged by hand into the ct-645 branch and can be closed.

PR #187, opened 2017-05-16
Rename open scope to opened to prevent clash with AR method

stephenrichards:


app/models/case.rb
```diff
@@ -253,6 +253,12 @@ def set_initial_state
     self.last_transitioned_at = Time.now
   end
 
+  def set_deadlines
```
misaka: Should this commit include deadlines changes?

app/services/deadline_calculator.rb
```diff
@@ -2,17 +2,17 @@ class DeadlineCalculator
 
   class << self
 
-    def escalation_deadline(kase)
+    def escalation_deadline(kase)   # press office deadline (not currently used)
```
misaka: Should this commit include deadline changes?

app/models/case.rb
stephenrichards: No it shouldn't.  Will remove

PR #191, opened 2017-05-19
CT-699 de-escalate case

misaka:
https://dsdmoj.atlassian.net/browse/CT-699

Add a `De-escalate` link to the incoming case list with an undo, similar to `Take case on`.


app/assets/javascripts/modules/DeEscalateCase.js
```diff
@@ -0,0 +1,33 @@
...
+  init: function () {
+    var hideShowEscalatedContainers = function (hideContainer, showContainer) {
+      return function() {
+        var $link = $(this)[0];
+        $.ajax({
+          url: $link.href,
+          method: 'patch',
+          statusCode: {
+            204: function() {
+              var $td = $($link.closest('td'));
```
aliuk2012: ```var td = $link.closest('td'); ```
You shouldn't need to wrap it inside jquery as its already a jQuery object.

misaka: I had an issue with `find` not being found on the object ... which is why I wrapped it.

app/assets/javascripts/modules/DeEscalateCase.js
```diff
@@ -0,0 +1,33 @@
+moj.Modules.CaseClosure = {
+  init: function () {
+    var hideShowEscalatedContainers = function (hideContainer, showContainer) {
+      return function() {
+        var $link = $(this)[0];
```
aliuk2012: Not sure why you converting `this` to a jQuery object only to convert it back to a native DOM.

app/assets/javascripts/modules/DeEscalateCase.js
misaka: Just tried it again, still getting:

```
TypeError: $td.find is not a function[Learn More]
```

app/assets/javascripts/modules/DeEscalateCase.js
misaka: I must be missing something ... I did this because without it, `$link.href` is undefined, breaking the `$.ajax` call. When I access the object like this it works ... `href` works on it and `closest('td')` still works so doesn't that mean that it's a jQuery object?

app/services/case_flag_for_clearance_service.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    managing_team_name = Settings.foi_cases.default_managing_team
+    managing_team = Team.managing.find_by(name: managing_team_name)
+    disclosure_team_name = Settings.foi_cases.default_clearance_team
+    approving_team = Team.approving.find_by(name: disclosure_team_name)
+    @case.state_machine.flag_for_clearance! @user,
+                                            managing_team,
+                                            approving_team
+    @case.approving_team = approving_team
+
```
stephenrichards: Don't you need to save @case here to record the approving team assignment?

spec/services/case_flag_for_clearance_service_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+        service.call
+        expect(assigned_case.state_machine)
+          .to have_received :flag_for_clearance!
+      end
+
+      it 'sets the result to ok and returns true' do
+        expect(service.call).to eq :ok
+        expect(service.result).to eq :ok
+      end
+    end
```
stephenrichards: I think this is the place where you should test that the expected action is actually taking place: i,e. the assignments are being set up on the case.

spec/services/case_unflag_for_clearance_service_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+        service.call
+        expect(assigned_flagged_case.state_machine)
+          .to have_received :unflag_for_clearance!
+      end
+
+      it 'sets the result to ok and returns true' do
+        expect(service.call).to eq :ok
+        expect(service.result).to eq :ok
+      end
+    end
```
stephenrichards: Similarly here, I think this is the place where  you should test that the approving assignment no longer exists.

app/services/case_flag_for_clearance_service.rb
misaka: Apparently not. I noticed that the save was extraneous while I was testing, I believe that because `@case.approving_team` is a relationship it doesn't need it.

spec/services/case_flag_for_clearance_service_spec.rb
misaka: Quite right, good spot thanks!

spec/services/case_unflag_for_clearance_service_spec.rb
misaka: And ditto.

spec/services/case_flag_for_clearance_service_spec.rb
misaka: Test added before this one.

spec/services/case_unflag_for_clearance_service_spec.rb
misaka: Test added before this one.

PR #194, opened 2017-05-22
CT-705 case details

aliuk2012:
https://dsdmoj.atlassian.net/browse/CT-705

Updates the Case details page layout.  As we were using a partial that were used on other pages I had to update those too. The accept/reject pages were updates and so was the close case page.


spec/views/cases/_case_status_html_slim_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+                   current_state: 'drafting',
+                   who_its_with: 'DACU'
+
+
+    render partial: 'cases/case_status.html.slim',
+           locals:{ case_details: unassigned_case}
+
+    cases_show_page.load(rendered)
+
+    partial = cases_show_page.case_status
```
misaka: Consider defining the a `case_status_section` in the SitePrism `application.rb` and using that here ... so that you'd be able to replace line 17-19 here with:

```
partial = case_status_section(rendered)
```

app/views/cases/show.html.slim
```diff
@@ -4,11 +4,67 @@
...
+          = action_button_for(event)
+
+  section.case-info
+    = render partial: 'cases/case_status', locals: {case_details: @case}
+
+
+  section.case-info
+    = render partial: 'cases/case_details', locals: {case_details: @case}
+
+  section.case-info
```
misaka: This section should probably be a partial since it's used by `assignments/edit.html.slim` and `assignments/show_rejected.html.slim`.

spec/site_prism/page_objects/pages/application.rb
```diff
@@ -24,7 +24,9 @@ def app_sections
         assignments_new:           'Assignments::NewPage',
         assignments_edit:          'Assignments::EditPage',
         global_nav_partial:        'Partials::GlobalNavPartial',
-        header_partial:            'Partials::HeaderPartial'
+        header_partial:            'Partials::HeaderPartial',
+        case_status_partial:       'Partials::CaseStatusPartial',
+        case_details_partial:      'Partials::CaseDetailsPartial',
```
misaka: Move the `header_partial`, `case_status_partial` and `case_details_partial` bits to the block below that creates sections (see `cases_what_do_you_want_to_do` for an example) and use these below instead of relying on the page tests to test them.

spec/site_prism/page_objects/sections/case_details_section.rb
```diff
@@ -0,0 +1,82 @@
+module PageObjects
+  module Sections
+    class CaseDetailsSection < SitePrism::Section
+
+      section :basic_details, '.basic-details' do
+        section :date_received, 'tr.date-received' do
+          element :heading, 'th'
```
misaka: You pointed out yourself that the `th`s aren't actually needed.

aliuk2012: Done

spec/views/cases/_case_status_html_slim_spec.rb
aliuk2012: Done

spec/site_prism/page_objects/pages/application.rb
aliuk2012: Done

app/views/cases/show.html.slim
aliuk2012: Done

spec/views/cases/_case_details_html_slim_spec.rb
```diff
@@ -57,27 +51,19 @@
       render partial: 'cases/case_details.html.slim',
              locals:{ case_details: unassigned_case}
 
-      cases_show_page.load(rendered)
-
-
-      partial = cases_show_page.case_details.basic_details
+      partial = case_details_section(rendered).basic_details
```
misaka: Next time we change this file, it'd make sense to use a less ambigous name for `partial` ... `basic_details` would work nicely, for example.

PR #199, opened 2017-05-23
CT-719 Approve Case

stephenrichards:
This PR does some of the background work to enable cases to be marked as approved, in
that it adds the approved boolean to assignments and checks that it can be true only
for approving assignments.


app/models/assignment.rb
```diff
@@ -2,19 +2,21 @@
...
+#  updated_at  :datetime         not null
+#  role        :enum
+#  user_id     :integer
+#  approved_at :datetime
 #
 
 class Assignment < ApplicationRecord
   validates :case, :role, :state, :team, presence: true
   validates :reasons_for_rejection, presence: true, if: -> { self.rejected? }
+  validate :approved_at_only_for_approvals
```
misaka: `approved_at` -> `approved`

app/models/assignment.rb
```diff
@@ -49,4 +51,12 @@ def assign_and_validate_state(state)
     assign_attributes(state: state)
     valid?
   end
+
+  private
+
+  def approved_at_only_for_approvals
```
misaka: `approved_at` -> `approved`

app/policies/case_policy.rb
```diff
@@ -17,6 +17,10 @@ def can_add_case?
     user.manager?
   end
 
+  def can_approve_case?
```
misaka: This probably shouldn't be empty.

PR #204, opened 2017-05-25
CT-696 in time and late tabs

misaka:
https://dsdmoj.atlassian.net/browse/CT-696

This implements tabs as an addition to the `GlobalNavManager` (changing `Entry` to `Page`), and restructures `CaseFinderService` to allow for additive filters so that tabs can add tab filters on-top of the page filter.


app/services/global_nav_manager.rb
```diff
@@ -1,48 +1,58 @@
...
-  attr_reader :nav_entries
-
-  def initialize(user)
-    @user = user
-    @nav_entries = []
-    add_views_for_user
+  def current_page
+    @current_page ||= Settings.global_navigation.pages.find do |_name, attrs|
+      url_for(attrs.path) == request.path
+    end&.tap { |name, _settings| break build_page name }
```
stephenrichards: I find this impossible to follow and understand.

PR #205, opened 2017-05-25
Add uniqueness validation for team name

stephenrichards:


app/models/assignment.rb
```diff
@@ -10,6 +10,7 @@
 #  updated_at :datetime         not null
 #  role       :enum
 #  user_id    :integer
+#  approved   :boolean          default(FALSE)
```
aliuk2012: Does this need to be here?

db/structure.sql
```diff
@@ -129,7 +129,8 @@ CREATE TABLE assignments (
     created_at timestamp without time zone NOT NULL,
     updated_at timestamp without time zone NOT NULL,
     role team_roles,
-    user_id integer
+    user_id integer,
+    approved boolean DEFAULT false
```
aliuk2012: Why are we adding "approved" to a unique team validation?

db/structure.sql
```diff
@@ -943,6 +944,7 @@ INSERT INTO "schema_migrations" (version) VALUES
 ('20170407091658'),
 ('20170420120713'),
 ('20170420122223'),
-('20170424133127');
+('20170424133127'),
+('20170523131602');
```
aliuk2012: If we dont need approved then this should go too

spec/factories/assignments.rb
```diff
@@ -10,6 +10,7 @@
 #  updated_at :datetime         not null
 #  role       :enum
 #  user_id    :integer
+#  approved   :boolean          default(FALSE)
```
aliuk2012: Does this need to be here?

spec/models/assignment_spec.rb
```diff
@@ -10,6 +10,7 @@
 #  updated_at :datetime         not null
 #  role       :enum
 #  user_id    :integer
+#  approved   :boolean          default(FALSE)
```
aliuk2012: Does this need to be here?

PR #207, opened 2017-05-26
cts script flagged cases

misaka:
https://dsdmoj.atlassian.net/browse/CT-741

A bunch of tweaks to the cts script like making validations more informative, using services for transitions that didn't, and allowing different case journeys. This change also changes the create command to be a sub-command of `cases`. `cts create` is now `cts cases create`.


misaka:
Putting into WIP as I need to update the README and the class structure.

PR #208, opened 2017-05-26
CT-719 clear response

stephenrichards:


app/policies/case_policy.rb
```diff
@@ -23,6 +23,10 @@ def can_add_case?
     user.manager?
   end
 
+  def can_approve_case?
```
misaka: I don't think we need this section, although I do like this name better than `can_approve?`.

PR #210, opened 2017-05-31
CT-711 "hide" long requests

aliuk2012:
https://dsdmoj.atlassian.net/browse/CT-711


app/helpers/cases_helper.rb
```diff
@@ -37,6 +37,23 @@ def action_button_for(event)
     end
   end
 
+
+  def show_hide_message(kase)
+
+    (part1,part2) = kase.message_extract
```
misaka: Variable naming here is confusing. Not fatally so, it's a short method, but there is two part 2s here, etc. How about, simply:

```
content_tag(:span, part1, class: 'ellipse-preview')
  + content_tag(:span,'...', class:'ellipse-delimiter js-hidden')
  + content_tag(:span, part2,  class: 'ellipse-complete js-hidden' )
  + link_to 'Show more', '#', class: 'ellipse-button js-hidden'
```

app/helpers/cases_helper.rb
```diff
@@ -40,17 +40,15 @@ def action_button_for(event)
...
-      part_1 = content_tag(:span, part1, class: 'ellipsis-preview')
-      part_2 = content_tag(:span,'...', class:'ellipsis-delimiter js-hidden')
-      part_3 = content_tag(:span, part2,  class: 'ellipsis-complete js-hidden' )
-      part_4 = link_to 'Show more', '#', class: 'ellipsis-button js-hidden'
-
-      part_1 + part_2 + part_3 + part_4
+      content_tag(:span, preview_copy, class: 'ellipsis-preview')
+        + content_tag(:span,'...', class:'ellipsis-delimiter js-hidden')
+        + content_tag(:span, remaining_copy,  class: 'ellipsis-complete js-hidden' )
+        + link_to 'Show more', '#', class: 'ellipsis-button js-hidden'
```
misaka: Sorry, looks like it needs parens around this one.

aliuk2012: I did think the part1+part2.... seemed rather messy I prefer your suggestion of removing them

PR #213, opened 2017-06-01
CT-733 Add Reassign-to me button

stephenrichards:


app/controllers/cases_controller.rb
```diff
@@ -120,6 +121,18 @@ def update
     end
   end
 
+  def reassign_approver
+    ars = ApproverReassignmentService.new(user: current_user, kase: @case)
+    if ars.call == :ok
+      flash[:notice] = 'Case re-assigned to you'
+      redirect_to case_path(@case)
+    else
+      flash[:error] = 'You do not have rights to re-assign this case to you'
```
misaka: Obviously "steal" would have sounded better here. :)

app/services/approver_reassignment_service.rb
```diff
@@ -0,0 +1,23 @@
+class ApproverReassignmentService
+
+  def initialize(user:, kase:)
+    @kase  = kase
+    @user = user
+    @policy = CasePolicy.new(user, kase)
```
misaka: We should probably stick with:

```
 Pundit.policy!(user, kase)
```

https://github.com/elabs/pundit#manually-retrieving-policies-and-scopes

lib/case_state_machine.rb
```diff
@@ -115,6 +115,16 @@ def self.get_policy(user_id, object)
     transition from: :pending_dacu_clearance, to: :awaiting_dispatch
   end
 
+  event :reassign_approver do
+    guard do |object, _last_transition, options|
+      CaseStateMachine.get_policy(options[:user_id], object).can_reassign_approver?
+    end
+
+    transition from: :awaiting_responder,     to: :awaiting_responder
```
misaka: Looking at which states we can perform this event, and also `accept_approver_assignment`, shouldn't we include `unassigned` since we can flag/unflag in that state too?

stephenrichards: We'd have to change accept_approver_assignment to transition from unassigned to unassigned for this to work - Cases should be only momentarily in unassigned state, I don't think it's worth it.

misaka: So why do we bother to allow `flag_for_clearance` in that state then? Seems to me the allowed states here should match. Happy to approve this through for now anyway.

PR #217, opened 2017-06-13
CT-754 press team

misaka:
https://dsdmoj.atlassian.net/browse/CT-754

Modifies the back-end to be able to handle multiple approvers. The focus of this change is just to be able to create cases that can be assigned multiple approvers (e.g. DACU Dislocure and Press Office) using the `cts` cmdline script without breaking existing features. Following tickets will add this functionality to the front-end, and will probably need some more tweaking to the back-end.

Here's some highlights of the changes here:

- addition of `-p` flag to `cts` to flag a case for Press Office
- general changes to controllers, services, etc so that they can query whether
  a case or assignment is with a given team or user, in addition to existing
  checks that only checked that the case/assignment had some approver
  assignments
- new model scopes and methods to support the above
- partial CasePolicy refactoring to encapsulate checks into named methods that
  are easier to parse, and that also support logging of failures for easier
  debugging. This refactoring is partial because I wanted to do as few changes
  here as possible (but, in the end, failed).
- addition of a `find_or_create` FactoryGirl strategy so that we can ensure that
  the DACU Disclosure team, for example, is a singleton (previously-attempted
  methods to do this were failing).


misaka:
There are some stylistic errors that have come, I'll fix them but they don't change the essence of the changes as they are.

cts
```diff
@@ -38,8 +77,49 @@ module CTS
...
+    def validate_data
+      validate_teams_populated
+      validate_users_populated
+      puts "Looks good."
+    end
+
+    private
+
+    def validate_teams_populated
+      @dacu_team, @disclosure_team, @hmcts_team, @hr_team, @laa_team = ['DACU', 'DACU Disclosure', 'Legal Aid Agency', 'HR', 'HMCTS North East Response Unit(RSU)'].map do |team_name|
```
aliuk2012: Should we not test Press Office?

cts
```diff
@@ -38,8 +77,49 @@ module CTS
...
+          teams.first
+        end
+      end
+    end
+
+    def validate_users_populated
+      begin
+        @dacu_manager = @dacu_team.managers.first ||
+                        raise("DACU BMT missing users")
+        @disclosure_approver = @disclosure_team.approvers.first ||
```
aliuk2012: Should we not include Press Office?

app/services/approver_reassignment_service.rb
```diff
@@ -9,8 +9,10 @@ def initialize(user:, kase:)
   def call
     if @policy.can_reassign_approver?
       ActiveRecord::Base.transaction do
-        assignment = @kase.approver_assignment
-        @kase.state_machine.reassign_approver!(@user, @kase.approver, @kase.approving_team)
+        assignment = @kase.approver_assignments.first
```
stephenrichards: Why are we just looking at the first approver_assignment? - aren't we in danger of reassigning an approver from dacu disclosure to a press office approver?  I would have thought that we should be getting the first approver assignment where the team is in the same team as the user?  Or have I misunderstood?

app/services/case_approval_service.rb
```diff
@@ -27,10 +27,10 @@ def able_to_approve?
 
   def mark_as_approved
     ActiveRecord::Base.transaction do
-      assignment = @kase.approver_assignment
+      assignment = @kase.approver_assignments.for_user(@user).first
```
stephenrichards: Same here = using first seems inherently dangerous

cts
stephenrichards: Agree with Al

app/models/case.rb
```diff
@@ -163,9 +165,13 @@ def self.search(term)
...
-  def awaiting_approver?
-    self.approver_assignment&.pending?
-  end
+  # Commented out as this is not being used and we don't know how to re-write
+  # this yet, we can wait until we actually need this method, and if we never
+  # do we should just delete it.
+  #
+  # def awaiting_approver?
+  #   self.approver_assignments.any? &:pending?
+  # end
```
misaka: This was meant to be, as you put it, an aide-memoir ... normally I'm not precious about leaving commented-out code around, but I thought this would help implementing this when we came to doing it later.

app/services/approver_reassignment_service.rb
misaka: I just wanted to limit the scope of the work here, doing this change would also mean having to change tests, etc, so I thought we'd wait until we're implementing this functionality from the front-end to fix it here.

app/services/case_approval_service.rb
misaka: Same as code above, I was just trying to limit the scope of this work, which is already quite large.

PR #221, opened 2017-06-15
CT-744 Allow Approvers to upload and return for redraft

stephenrichards:


app/helpers/cases_helper.rb
```diff
@@ -35,9 +35,14 @@ def action_button_for(event)
               class: 'button'
     when :upload_response_and_approve
       link_to t('common.case.upload_approve'),
-              new_response_upload_case_path(@case, 'action' => 'upload-approve'),      # upload and clear
+              new_response_upload_case_path(@case, 'action' => 'upload-approve'),
               id: 'action--upload-approve',
               class: 'button'
+    when :upload_response_and_return_for_redraft
+      link_to t('common.case.upload_and_redraft'),
+              new_response_upload_case_path(@case, 'action' => 'upload-revert'),
```
misaka: Would it make more sense to call this action `upload-redraft`? Here and in `response_uploader_service.rb`?

stephenrichards: But we could do with a bit more consistency it's true.  We've got add responses for some events, and upload for others....

spec/features/cases/foi/disclosure_specialist_approval_spec.rb
```diff
@@ -158,4 +162,26 @@ def upload_response_with_action_param(kase, user, action)
     expect(cases_show_page.case_status.details.copy.text).to eq 'Ready to send'
     expect(cases_show_page.case_status.details.who_its_with.text).to eq responding_team.name
   end
+
+  scenario 'upload a response and return for redraft', js: true do
+    kase = create_flagged_case_and_assign_to_team
```
misaka: I thought we were going to stop repeatedly exercising this part of the journey (case creation et al) and use factories?

app/helpers/cases_helper.rb
misaka: :+1:

spec/features/cases/foi/disclosure_specialist_approval_spec.rb
```diff
@@ -158,4 +162,26 @@ def upload_response_with_action_param(kase, user, action)
...
+    accept_case_as_kilo(kase)
+    upload_response_as_kilo(kase.reload, responder)
+
+    login_as disclosure_specialist
+    take_case_on_as_discosure_specialist(kase)
+    cases_show_page.load(id: kase.id)
+    expect(cases_show_page.actions).to have_upload_redraft
+    cases_show_page.actions.upload_approve.click
```
misaka: Is this right? Shouldn't you be clicking `upload_redraft` here?

And if so, you probably won't need `expect(cases_show_page.actions).to have_upload_redraft` since that's tested when you try to click on it.

PR #226, opened 2017-06-21
CT-764 group attachments

stephenrichards:
Present attachments to the view in a collection of Upload groups so that they can be easitly itereated through on the page.


db/migrate/20170609094110_add_group_and_user_to_case_attachments.rb
```diff
@@ -0,0 +1,15 @@
+class AddGroupAndUserToCaseAttachments < ActiveRecord::Migration[5.0]
+
+  def up
+    add_column :case_attachments, :upload_group, :string
+    add_column :case_attachments, :user_id, :integer
+
+    require File.join(Rails.root, 'db', 'seeders', 'case_attachment_upload_group_seeder')
+    CaseAttchmentUploadGroupSeeder.new.run
```
misaka: Missing the second 'a' in `Attchment`. :)

spec/features/cases/foi/case_viewing/removing_response_spec.rb
```diff
@@ -70,8 +70,8 @@
       end
 
       context 'when there are multiple responses' do
-        background do
-          other_response = create :case_response, case: case_with_response
+        before do
```
misaka: You can revert this back to `background` now.

spec/features/cases/foi/case_viewing/removing_response_spec.rb
```diff
@@ -152,8 +152,8 @@
       end
 
       context 'when there are multiple responses' do
-        background do
```
misaka: Also could be reverted to `background`, although clearly it'll still work.

spec/views/cases/_case_attachments_html_slim_spec.rb
```diff
@@ -0,0 +1,67 @@
+# require 'rails_helper'
```
misaka: Does this need to be in the PR?

PR #233, opened 2017-06-23
CT-805 Display upload group time in local time zone

stephenrichards:


app/models/case_attachment_upload_group.rb
```diff
@@ -37,4 +37,13 @@ def delete!(case_attachment_id)
...
     other.timestamp <=> @timestamp
   end
+
+  private
+  def get_time(upload_group)
+    current_timezone = Time.zone
+    Time.zone = 'Etc/UTC'
+    t =  Time.zone.parse(upload_group)
+    Time.zone = current_timezone
+    t.in_time_zone
```
misaka: I think this can be abbreviated as:

```
Time.find_zone('Etc/UTC').parse(upload_group).in_time_zone
```

PR #240, opened 2017-06-28
CT-807 Copy changes

stephenrichards:


aliuk2012:
If you create a trigger case and assign it to a KILO, logging as the KILO and accept the case. Click on "Upload response". Then do not upload any files when you on the Upload response page and just click on the action button.

image

PR #246, opened 2017-06-30
Add further command to cts to enable demo data to be produced

stephenrichards:
The following command:
```
    cts cases demo -n100
```

Will generate 100 cases in a variety of states with a variety of received dates,
and (if appropriate) responded dates.


lib/cts/demo_setup.rb
```diff
@@ -0,0 +1,137 @@
...
+
+module CTS
+  class DemoSetup
+
+    include FactoryGirl::Syntax::Methods
+
+    def initialize(number)
+      @number = number
+      @cases = Case.order(id: :desc).limit(number).to_a
+      @managing_team = Team.where(name: 'DACU').first
```
misaka: Just for reference, `CTS::dacu_team` is a short-cut that standardises finding the DACU team.

PR #248, opened 2017-07-03
CT-798 Display attachments for postal FOI

aliuk2012:
https://dsdmoj.atlassian.net/browse/CT-798

I had to apply a filter to the cases_helpers.rb
case_attachments_visible_for_case? method to filter out requests in the
case details view.

I renamed the Case models upload_groups method to upload_response_groups
and then filtered out the attachments before calling the
CaseAttachmentUploadGroupCollection. This meant that we could reuse the
collection without changing it. I create an upload_request_groups method
in line with upload_response_groups.


misaka:
I'm just waiting for CT-767 to be merged so this can be re-based properly off of it.

app/views/cases/_case_request.html.slim
```diff
@@ -3,9 +3,33 @@
     = t('common.case.request')
 
   .request--message
-    p
-      - if defined?(full_message) && full_message
-        = case_details.message
-      - else
-        = show_hide_message(case_details)
+    - if case_details.delivery_method == 'sent_by_email'
```
misaka: Just to say, the reason we used the `sent_by_email` wording is so that we can do: `case_details.sent_by_email?` here. :)

app/views/cases/_case_request.html.slim
```diff
@@ -3,9 +3,33 @@
...
+      p
+        - if defined?(full_message) && full_message
+          = case_details.message
+        - else
+          = show_hide_message(case_details)
+    - else
+      - case_details.upload_request_groups.each do |ug|
+        .case-attachments-group
+          h3.case-attachments--primary
+            = ug.date_time
```
misaka: Just remembering the issues we had elsewhere, will this be in the correct timezone?

app/models/case.rb
```diff
@@ -214,8 +214,12 @@ def self.search(term)
   end
 
 
-  def upload_groups
-    CaseAttachmentUploadGroupCollection.new(self, attachments)
+  def upload_response_groups
```
misaka: We should have tests for this and `upload_request_groups` below.

app/views/cases/_case_request.html.slim
aliuk2012: It should do because its handled in the CaseAttachmentUploadGroup model

https://github.com/ministryofjustice/correspondence_tool_staff/blob/develop/app/models/case_attachment_upload_group.rb#L17

misaka: 👍 I guess it's easily confirmed seeing as we're currently in BST.

stephenrichards: Yes - the upload group takes care of putting it in the right timezone

PR #253, opened 2017-07-06
CT-823 open foi stats

stephenrichards:
Incorporates CT-825 Refactor who_its_with method on Case, CaseDecorator


app/decorators/case_decorator.rb
```diff
@@ -1,13 +1,15 @@
 class CaseDecorator < Draper::Decorator
   delegate_all
+  #
+  # if the current user is in the same team as the team name, display the user
+# else display team name
```
misaka: Not sure this comment is helping, formatting makes it unclear it's for the method below and wording isn't clear ... maybe something along the lines of:

```
  # if the case is with a responding team and the current user is a responder
  # in that team, display the name of the specific user it's with instead of
  # the team name
```

PR #256, opened 2017-07-10
CT-828 Implement R003 Current Month Case Overview Report

stephenrichards:


app/services/stats/base_report.rb
```diff
@@ -0,0 +1,28 @@
+module Stats
+  class BaseReport
+
+
+    def initialize
+      raise "Cannot instantiate Stats::BaseReport - use derived class instead" if self.class == BaseReport
```
misaka: Shouldn't this be a mixin, if we never want to derive from it?

PR #263, opened 2017-07-13
CT-844 Add new outcome Request Clarification

stephenrichards:


db/migrate/20170713094438_add_request_clarification_ooutcome.rb
```diff
@@ -0,0 +1,9 @@
+class AddRequestClarificationOoutcome < ActiveRecord::Migration[5.0]
```
misaka: Is this a distant relative to the Oort Cloud? ;)

stephenrichards: yep
:-)

stephenrichards: It's only the name of a  migration, so let's leave it for now and get it in.

PR #266, opened 2017-07-14
CT-837 Cabinet Office Report: Timeliness and Outcomes sections

stephenrichards:


app/models/case_closure/exemption.rb
```diff
@@ -1,4 +1,4 @@
-# == Schema Information
+# == Schema Inforails rmation
```
misaka: Is this `Info(rails )rmation` or `Infor(ails r)mation`? :)

PR #267, opened 2017-07-14
CT-812 add guards to state machine event transitions

misaka:
Initial commit to add functionality off having per-transition guards.


spec/lib/case_state_machine_spec.rb
```diff
@@ -177,15 +177,19 @@ def event(event_name)
 
     it { should transition_from(:pending_dacu_clearance)
                   .to :pending_press_office_clearance }
-    it { should require_permission(:can_escalate_to_next_approval_level?)
+    xit { should require_permission(:can_escalate_to_next_approval_level?)
```
aliuk2012: xit !!

spec/lib/case_state_machine_spec.rb
```diff
@@ -177,15 +177,19 @@ def event(event_name)
...
-    it { should transition_from(:pending_dacu_clearance).to :awaiting_dispatch}
-    it { should transition_from(:pending_press_office_clearance).to :awaiting_dispatch}
-    it { should require_permission(:can_approve_case?)
+    it { should transition_from(:pending_dacu_clearance)
+                  .to(:awaiting_dispatch)
+                  .checking_case_policy(:can_approve_case?) }
+    it { should transition_from(:pending_press_office_clearance)
+                  .to(:awaiting_dispatch)
+                  .checking_case_policy(:can_approve_case?) }
+    xit { should require_permission(:can_approve_case?)
```
aliuk2012: same as above

spec/site_prism/page_objects/pages/cases/show_page.rb
```diff
@@ -8,6 +8,7 @@ class ShowPage < SitePrism::Page
                 PageObjects::Sections::PrimaryNavigationSection, '.global-nav'
 
         element :escalation_notice, '.alert-orange'
+        elements :notices, '.notice-summary-heading'
```
aliuk2012: shouldn't this be `element`? i.e only expect on notice?

misaka: Ya, I was thrown off by the `.each` I showed you, but you're right ... it should only ever get one of those keys. :100:

aliuk2012:
@misaka code coverage has now dropped down by 22.4%

misaka:
Presumably the code coverage dropping is because of the error.

aliuk2012:
Tests are passing now but code coverage still dropped 22%

misaka:
So why does https://codeclimate.com/github/ministryofjustice/correspondence_tool_staff show 98% \*scratching-head\*

PR #270, opened 2017-07-19
CT-811 private office take on case

misaka:
https://dsdmoj.atlassian.net/browse/CT-811

Allows taking on the case by Private Office, along with associated teams. This change will require an env var `SETTINGS__PRESS_OFFICE_DEFAULT_USER` in production, and will probably require a little more testing.


app/services/case_flag_for_clearance_service.rb
```diff
@@ -14,10 +14,10 @@ def call
 
     if @team.dacu_disclosure?
       assign_approver(@user, @team, @team)
-    elsif @team.press_office?
-      assign_and_accept_approver(@user, @team)
-      if @case.approver_assignments.where(team: @dts.approving_team).blank?
-        assign_approver(@user, @team, @dts.approving_team)
+    elsif @team.press_office? || @team.private_office?
+      assign_and_accept_approver(@user, @team, @team)
```
aliuk2012: double @team seems weird is this correct?

misaka: I can add argument keywords if that helps.

aliuk2012: I think that would otherwise, you have to think about what that second argument is

PR #271, opened 2017-07-19
CT-806 open cases page for private office

misaka:
https://dsdmoj.atlassian.net/browse/CT-806

This functionality is already present with the existing code. The Take Case On feature test has been changed to include testing that the open cases page works, but other than that there isn't much more to add, there aren't even team-specific view tests to change.


spec/features/cases/foi/case_approvals/private_office_approvals_spec.rb
```diff
@@ -30,10 +30,17 @@
...
     case_row.actions.take_on_case.click
     expect(case_row.actions.success_message).to have_text 'Case taken on'
+
     case_row.number.click
     expect(cases_show_page.case_history.entries.first)
       .to have_text('Preston Offman Take on for approval')
     expect(cases_show_page.case_history.entries[1])
       .to have_text("#{private_officer.full_name} Flag for clearance")
+
+    pending_dacu_clearance_case
```
aliuk2012: why are we using pending_dacu_clearnance_case for a private office test?

misaka: I just wanted to populate the DB with a case that shouldn't show up on the open cases page ... and I obviously don't want to go and create cases in every possible state for testing.

aliuk2012: Sure, I understand that you wouldn't want to create a case for every possible state in testing. I was merely asking encase the spec was giving a false positive result because the wrong case might be used.

PR #273, opened 2017-07-21
CT-819 allow private office to request amends

misaka:
This is the counter-point to press/private office case approvals and as such
looks very similar in implementation. We can likely refactor so that
approver/request amends can re-use some of the same bits, such as the view which
nearly identical.


misaka:
Updated to remove an `fscenario`.

misaka:
And ... wrong button. why why why is that button right next to "Comment" there??

PR #279, opened 2017-07-26
CT-861 - Fix state transitions for Unflag for Clearance event

stephenrichards:


app/policies/case_policy.rb
```diff
@@ -189,6 +189,46 @@ def new_response_upload?
     check_user_is_in_current_team
   end
 
+  def unflag_for_clearance_from_unassigned_to_unassigned?
+    clear_failed_checks
+    check_user_is_dacu_disclosure_approver ||
+      check_user_is_assigned_approver_for_case ||
+      (check_user_is_a_manager && check_case_requires_clearance)
```
misaka: I think that `check_user_is_dacu_disclosure_approve` has to also `check_case_requires_clearance` because it only tests that the user is on the DACU Disclosure team. i.e.

```
   (check_user_is_dacu_disclosure_approver && check_case_requires_clearance) ||
     check_user_is_assigned_approver_for_case ||
     (check_user_is_a_manager && check_case_requires_clearance)
```

But perhaps dried up.

app/policies/case_policy.rb
```diff
@@ -189,6 +189,46 @@ def new_response_upload?
...
+  end
+
+  def unflag_for_clearance_from_pending_dacu_clearance_to_awaiting_dispatch?
+    clear_failed_checks
+    (check_user_is_a_manager && check_case_requires_clearance) ||
+      (check_user_is_an_approver_for_case && check_case_has_no_other_approving_teams)
```
misaka: For this transition we need to `check_case_has_no_other_approving_teams` in all cases, even when the user is a manager.

app/policies/case_policy.rb
```diff
@@ -189,6 +189,46 @@ def new_response_upload?
...
+  end
+
+  def unflag_for_clearance_from_pending_dacu_clearance_to_pending_dacu_clearance?
+    clear_failed_checks
+    check_user_is_an_approver_for_case && check_case_has_another_approving_team
```
misaka: I think DACU Disclosure team members and managers should be able to do this, assuming the case requires clearance of course.

app/policies/case_policy.rb
```diff
@@ -197,6 +197,69 @@ def new_response_upload?
...
+        check_user_is_an_approver_for_case
+    ) ||
+    (
+      check_case_requires_clearance_from_dacu_disclosure_and_another_team &&
+        (
+          check_user_is_a_manager ||
+            check_user_is_dacu_disclosure_approver
+        )
+    )
+  end
```
misaka: Can we simplify this like so:

```
(check_case_is_assigned_to_press_office && check_user_is_press_office_approver) ||
  (check_case_is_assigned_to_private_office && check_user_is_private_office_approver)
```

Does it make sense for DACU Disclosure or Managers to unflag a case that is assigned to Press of Private? If so we'll probably need a different transition to go `pending_dacu_clearance -> pending_press_office_clearance`.

app/policies/case_policy.rb
```diff
@@ -197,6 +197,69 @@ def new_response_upload?
...
+  def unflag_for_clearance_from_pending_dacu_clearance_to_awaiting_dispatch?
+    clear_failed_checks
+    check_case_requires_clearance &&
+    (
+      check_user_is_a_manager ||
+        ( check_case_has_no_other_approving_teams &&
+          check_user_is_an_approver_for_case
+        )
+    )
+  end
```
misaka: Shouldn't only DACU Disclosure and DACU BMT be able to do this transition? How about:

```
check_case_requires_clearance && check_case_has_no_other_approving_teams &&
  (check_user_is_a_manager || check_user_is_dacu_disclosure_approver)
```

spec/models/business_unit_spec.rb
```diff
@@ -62,14 +62,15 @@
...
+                                            BusinessUnit.dacu_disclosure,
                                             approving_team
                                           ]
       end
     end
   end
 
   it 'has a working factory' do
-    expect(create :business_unit).to be_valid
+    expect(find_or_create :business_unit).to be_valid
```
misaka: In the `business_unit` factory, we're not setting the email address to a pre-determined value ([see above](https://github.com/ministryofjustice/correspondence_tool_staff/pull/279/files#diff-d3de95f57f07da1a2820052964371e07R4)), so `find_or_create` is unlikely to have an affect here. Similarly with `managing_team`, `responding_team` and `approving_team` above. Can we actually be sure that this is actually required? Just so we don't start reciting useless incantations throughout the code.

PR #281, opened 2017-07-27
CT-789 Prevent responder from uploading response until escalation dea…

stephenrichards:
…dline has passed

This change removes the upload response button until the escalation deadline has passed.
It also changes the escalation deadline from 6 days to 3 days.


spec/controllers/cases_controller_spec.rb
```diff
@@ -489,6 +489,7 @@ def case_closure_params(kase)
             .to eq [:add_message_to_case,
                     :add_response_to_flagged_case,
                     :reassign_user]
+        expect(assigns(:permitted_events)).to eq [:add_message_to_case, :add_response_to_flagged_case, :reassign_user]
```
misaka: Duplication with the expect above I believe.

spec/features/cases/foi/case_approvals/disclosure_specialist_approval_spec.rb
```diff
@@ -1,5 +1,7 @@
 require 'rails_helper'
 
+include CaseDateManipulation
```
misaka: Weren't we going to see if we can have this automatically included in `rails_helper.rb`?

PR #282, opened 2017-07-27
CT-846 keep action_params when uploaded files are blank

misaka:
This fixes a 500 error caused when the user accidentally tries to upload an
empty list of files.


spec/controllers/cases_controller_spec.rb
```diff
@@ -1000,20 +1000,40 @@ def do_upload_responses
...
+        end
       end
 
-      it 're-renders the page if there is an upload error' do
-        allow_any_instance_of(CasesController).to receive(:flash).and_return(flash)
-        expect(ResponseUploaderService).to receive(:new).and_return(response_uploader)
-        expect(response_uploader).to receive(:result).and_return(:error)
-        do_upload_responses
-        expect(response).to have_rendered(:new_response_upload)
+      fcontext 'there is an upload error' do
```
aliuk2012: Fcontext.....

PR #285, opened 2017-07-27
CT-859 Create TeamProperties to flag teams which are allocatable by C…

stephenrichards:
…ategory

This PR adds the team_proerties table which allows us to flag which teams can have cases of which category allocated to them, as well as specify name of the Team Leader, Areas covered, etc.


db/migrate/20170727112001_create_team_properties.rb
```diff
@@ -0,0 +1,10 @@
+class CreateTeamProperties < ActiveRecord::Migration[5.0]
+  def change
+    create_table :team_properties do |t|
+      t.integer :team_id
```
misaka: We should likely index this column

app/models/team_property.rb
```diff
@@ -0,0 +1,18 @@
+class TeamProperty < ActiveRecord::Base
+
+
+  VALID_KEYS = %w(
+    area
+    lead
+    can_allocate
+  )
+
+  validates :value, uniqueness: { scope: [:team_id, :key], message: "%{value} is not unique in team and key" }
```
misaka: Is it really important this is unique? Is it just a safeguard against accidental duplication?

stephenrichards: yes, I  think it is important that it's unique - we don't want duplication.  But, thinking about it, validates_uniqueness is notoriously slow, so it might be better to put a unique index on the three columns.

PR #294, opened 2017-08-02
CT-883 send assignment notification to all users on team

misaka:
When a team doesn't have a group email, then we notify the responders on that
team by sending an email to each one of them individually.


misaka:
The failed test is the sporadic error we get in the upload feature test. Not related to my changes. Merging.

PR #295, opened 2017-08-02
CT-869 show business groups on allocation page

aliuk2012:
Displays a list of business groups on the allocation page and allows the user to filter business units by business groups


app/views/assignments/new.html.slim
```diff
@@ -9,12 +9,29 @@
...
+    - @business_units.each do |team|
+      li.team
+        .team-details
+          h3
+            .team-unit-name
+              = team.name
+        .team-actions
+          = link_to t('button.assign'),
+                  case_assign_to_responder_team_path(case_id: @case.id,
+                          team_id: team.id), class: 'button'
```
misaka: My understanding is that the buttons would be too noisy, visually, and we were going to use a link instead of a button. Has the design changed or are you challenging it? :)

spec/features/cases/foi/case_assigning_spec.rb
```diff
@@ -22,7 +22,12 @@
     click_link 'Assign to a responder'
     expect(assignments_new_page).to be_displayed
 
-    assignments_new_page.assign_to.team.first.assign_link.click
+    # Browse Business Group
+    assignments_new_page.choose_business_group(responder.responding_teams.first
+                                                   .business_group)
+
+    # Select Business Unit
+    assignments_new_page.choose_business_unit(responder.responding_teams.first)
```
misaka: Is this the same thing as what's in the `assign_case_to_team` method defined in `disclosure_specialist_approval_spec.rb`? How about moving it into `spec/support/feature_steps` and re-using it here and in the specs below?

app/views/assignments/new.html.slim
aliuk2012: Mara, Kellie and myself had a chat about it and we agreed to go with "buttons" as it is an action that they are taking on the case. Also a link on its own doesn't make it clear exactly what it is the user is expected to do next. Also once we have searching in place there shouldn't really be a large pool of units to select from

spec/features/cases/foi/case_approvals/disclosure_specialist_approval_spec.rb
```diff
@@ -22,7 +22,14 @@ def create_case(flag_for_clearance: false)
...
 
-    assignments_new_page.assign_to.team.first.assign_link.click
+    # Browse Business Group
+    assignments_new_page.choose_business_group(responder.responding_teams.first
+                                                  .business_group)
+
+    # Select Business Unit
+    assignments_new_page.choose_business_unit(responder.responding_teams.first)
+
+    expect(cases_show_page.case_status.details.copy.text).to eq "To be accepted"
```
misaka: Would it make sense to test that the correct assignment is done here? i.e., the case is now "with" the team that was selected?

app/views/assignments/new.html.slim
misaka: 👍 ... I liked the button better anyway. :)

spec/features/cases/foi/case_approvals/disclosure_specialist_approval_spec.rb
aliuk2012: I'll do this as part of the other refactoring suggestion we spoke about

PR #300, opened 2017-08-02
CT-867 show 'Areas covered' and 'Deputy director'

aliuk2012:
This covered both CT-867 and CT-870 which talks about displaying the 'Areas covered' and 'Deputy Directors' details to DACU BMT when they are assigning a case.

This also has two side commits as follows:
- Update nokogiri due to security vulnerablity that Code Climat had raised
- Refactor assignment spec specs that @misaka and I spoke about under #295


spec/views/assignments/new_page_spec.rb
```diff
@@ -53,6 +53,11 @@
 
       expect(page.assign_to.team.count).to eq 3
 
+      page.assign_to.team.each do | team |
+        expect(team).to have_areas_covered
+        expect(team).to have_deputy_director
```
misaka: I think we should probably check to make sure that these are actually being populated with data from the team.

spec/views/assignments/new_page_spec.rb
```diff
@@ -53,10 +54,22 @@
...
       expect(page.assign_to.team.count).to eq 3
 
-      any_team = page.assign_to.team.first
-      expect(any_team.business_unit.text).to eq business_unit_1.name
-      expect(any_team.assign_link.text).to eq "Assign to this unit"
-      expect(any_team.assign_link[:href]).to eq case_assign_to_responder_team_path(unassigned_case, team_id: business_unit_1.id )
+      all_business_units.each do | bu |
+        page_team = page.assign_to.team.detect { |team| team.business_unit.text == bu.name }
+
+        expect(page_team.business_unit.text).to eq bu.name
```
misaka: This is already covered by line 58 ... there's no way for this to fail.

aliuk2012: :( But I was really hoping that it would fail the second time round

PR #305, opened 2017-08-04
CT-922 Minor copy changes

stephenrichards:


config/locales/en.yml
```diff
@@ -118,7 +118,7 @@ en:
         requester_type: Type of requester
         refusal_reason_name: What was the reason for refusal?
         outcome_name: What was the outcome?
-        flag_for_disclosure_specialists: Does a disclosure specialist need to see this case?
+        flag_for_disclosure_specialists: Clearance required by DACU Disclosure?
```
aliuk2012: Shouldn't we use SETTINGS instead of hard coding a teams name in because if the name needs to change we will have to remember all the places we have used it

config/locales/en.yml
```diff
@@ -142,7 +142,7 @@ en:
         message: Full request
         postal_address: Address
         flag_for_disclosure_specialists:
-          Does a disclosure specialist need to see this?
+          Clearance required by DACU Disclosure?
```
aliuk2012: Shouldn't we use SETTINGS instead of hard coding a teams name in because if the name needs to change we will have to remember all the places we have used it

config/locales/en.yml
```diff
@@ -152,8 +152,8 @@ en:
       <strong>You've accepted this case</strong><br>
       It will now appear in <a href="%{case_list_url}">your cases</a>.
     assignment_rejected_html: |
-      <strong>Your response has been sent</strong><br>
-      This case will be reviewed and assigned the to appropriate unit.
+      <strong>You've rejected this case</strong><br>
+      DACU BMT will assign the case to the appropriate business unit.
```
aliuk2012: Do our users know who DACU BMT are? I think we have come up with the terms "DACU BMT" and "DACU Disclosure Specialists" team name so its clear for us to take about the two different roles within DACU. Our users might only know DACU as DACU and not be aware of sub sections and I dont think we should hard code in team names

config/locales/en.yml
misaka: My feeling is we can put this into `en.yml`, but it should either be in a separate section for FOIs, or maybe we'll create a whole new file ... but until we need to split this out, we may be able to not worry about it.

config/locales/en.yml
misaka: I think we need to feed these back to Kellie & Mara, they won't see these comments here.

kelliematheson: Names of the teams are still being decided by DACU. For now it is fine to have the current names hardcoded, this shouldn't block merge and release of this ticket.

config/locales/en.yml
misaka: Having had another look at these, I now think:

1. you're right, we should actually parameterise DACU Disclosure's name, but ...
2. we can't do this now, as per @stephenrichards's comments on https://dsdmoj.atlassian.net/wiki/spaces/CD/pages/144219662/Design+-+Sprint+26+-+Small+copy+changes

Let's worry about this when we have to, if we have to change the name it'll be slightly annoying, but nothing a little bit or Ruby can't help us with:

```
ruby -i.bak -pe '$_.gsub!(/DACU Disclosure/, "DACU are hosers")' **/*.rb
```

PR #313, opened 2017-08-14
CT-922 minor changes from QA tests

aliuk2012:
Fixes the issues raised in QA testing from CT-922

*QA TEST*
Tested on dev.
- Case detail page - should be 'Date received at MOJ'
- My open cases - heading should be 'My open cases'
- Case creation - Remove question mark on 'Clearance required by DACU - Disclosure'


app/controllers/cases_controller.rb
```diff
@@ -30,6 +30,7 @@ def index
                 .cases
                 .page(params[:page])
                 .decorate
+    @current_tab_name = 'all_cases'
```
misaka: Adding stuff to the controller seems a bit messy to me. Can we use a helper in the view and access the name of the action with `action_name` instead?

misaka: Just checked and `action_name` works from inside the view and returns the name of the action.

PR #314, opened 2017-08-15
CT-709 Display clearance levels on case detail page

stephenrichards:


app/decorators/case_decorator.rb
```diff
@@ -85,6 +89,18 @@ def self.collection_decorator_class
     PaginatingDecorator
   end
 
+  def responding_team_lead_name
+    object.responding_team&.team_lead&.value
```
misaka: Do we want to display an empty string when there is no responding team, or if there is no team lead, etc? Don't we want it to blow up?

misaka: Just spoke to Kellie and she confirms that the Disclosure Specialist line should show up, but with no Disclosure Specialist, if they haven't accepted the approver assignment by day 4, so this code is correct.

PR #319, opened 2017-08-18
CT-825 Normalize CaseTransition team and user_ids

stephenrichards:
Case transitions used a number of fields in the metdata to describe
who was carrying out the action and whoc was the recipient.  The change normalizes
those fields, and moves them into attributes on the database.  The new fields are:
 - acting_user_id
 - acting_team_id
 - target_user_id
 - target_team_id


db/migrate/20170818082409_add_user_team_attrs_to_case_transitions.rb
```diff
@@ -0,0 +1,9 @@
+class AddUserTeamAttrsToCaseTransitions < ActiveRecord::Migration[5.0]
+  def change
+    add_column :case_transitions, :acting_user_id, :integer
```
misaka: We do the occasional lookups on these, should we index any of them? For example, the case unassigment service does a lookup on `acting_team_id` and `target_team_id`, although in that example the search would be scoped to the case so it's probably not an issue. Are there any queries we do across all cases that might want an index? In stats?

PR #332, opened 2017-08-25
Allow Add / Updates to all team  members regardless of team role

stephenrichards:


app/views/teams/_business_unit_detail.html.slim
```diff
@@ -36,7 +36,7 @@ p
 - if Pundit.policy(current_user, team).can_add_new_responder?
```
colinbruce: Has the pundit policy been update already?

PR #341, opened 2017-09-05
CT-967 Datasbase dump and reload rake tasks

stephenrichards:
This PR implements rake tasks that create an SQL dump of the
database on production, demo or staging environments and copy
to the localhost, and to load that SQL dump to the local database.

rake tasks are:

```
    rake db:dump:prod
    rake db:dump:staging
    rake db:dump:demo
    rake db:load:local
```

Still to do:

 - Anonymise the case data
 - Allow loading to dev, demo, staging environments


Gemfile
```diff
@@ -68,16 +69,20 @@ group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   gem 'annotate'
   gem 'byebug', platform: :mri
+  gem 'colorize'
   gem 'guard-jasmine'
   gem 'launchy'
   gem 'mechanize', '~> 2.7.5'
+  gem 'net-ssh'
+  gem 'net-scp'
```
misaka: If you're satisfied with using the ssh binary, you can remove these two.

lib/db/database_dumper.rb
```diff
@@ -0,0 +1,90 @@
+require 'net/ssh' # gem install net-ssh
+require 'net/scp' # gem install net-scp
```
misaka: If you're satisfied with using the ssh binary, you can remove these two.

Gemfile
```diff
@@ -68,16 +69,20 @@ group :development, :test do
...
   gem 'parallel_tests'
   gem 'pry'
   gem 'pry-byebug'
   gem 'rb-readline', '~> 0.5.4'
   gem 'rspec-rails', '~> 3.4'
   gem 'rubocop', require: false
   gem 'rubocop-rspec', require: false
+  gem 'ruby-progressbar'
```
misaka: I don't see this being used anywhere.

lib/db/database_dumper.rb
```diff
@@ -0,0 +1,90 @@
...
+require 'colorize'
+require 'shell-spinner'
+require 'ruby-progressbar'
```
misaka: Unused?

lib/tasks/db_dump.rake
```diff
@@ -0,0 +1,65 @@
...
+      puts ' specified environment (prod, demo, staging).'.yellow
+      puts ' '
+      puts 'SSH username:'.yellow
+      puts 'The ssh username used to connect to the remote server must be specified in the'.yellow
+      puts 'CTS_SSH_USER environment variable.'.yellow
+      puts ' '
+      puts 'Hostname:'.yellow
+      puts 'The hostname or IP address can be specified as a parameter, for example'.yellow
+      puts ' '
+      puts '   rake db:dump:demo[123.202.5.66]'.yellow
```
misaka: How about:

```
  rake db:dump:demo[user@123.202.5.66]
```

So that there's no need to set `CTS_SSH_USER` env variable?

stephenrichards: because I like to do it with a few params as possible.

stephenrichards: lets get this in so that at least we have it working, and then I'll iterate to make that an option when I add in the anonymising.

misaka: Will you remove `ruby-progressbar` at the same time then?

PR #349, opened 2017-09-07
CT-1043 Service does not support IE7/8/9

aliuk2012:
Our users who have older versions of IE 7/8/9 should have access to
other modern browsers. This adds a banner to inform them that the
service does not support their older browser.


app/views/devise/sessions/new.html.slim
```diff
@@ -1,3 +1,13 @@
+/[if lte IE 9]
+  .error-summary(role="group" aria-labelledby="old-browser" tabindex='-1')
+    h1.heading-medium.notice-summary-heading(id= "old-browser")
+      | This service may not work using this version of Internet Explorer
+    p
+      | Please use another web browser (for example Firefox).
```
misaka: Should we be more specific? Like, "Please use Firefox" since that's what we expect them to use? Should there be a link?

PR #351, opened 2017-09-11
CT-1004 adds a migration to add codes to teams with a default value o…

beccapearce:
…f nil

updates the default_managing_team and default_clearance_team within settings to use code andchanges searches for dacu to use the code state, breaks many things

changes instance using default_managing_team and default_clearance_team to the new default_managing_team_name and default_clearance_team_name

wip

updates seeds to use team code and removes methods searching for teams by name in favour of searching by code methods'

adds code to global nav


app/models/business_unit.rb
```diff
@@ -49,11 +49,19 @@ def valid_role
...
-    find_by!(name: Settings.foi_cases.default_clearance_team)
+    find_by!(code: Settings.foi_cases.default_clearance_team)
   end
 
   def dacu_disclosure?
-    name == Settings.foi_cases.default_clearance_team
+    code == Settings.foi_cases.default_clearance_team
+  end
+
+  def self.dacu_bmt
```
misaka: Let's make sure we have tests for this method and `dacu_bmt?` below.

db/structure.sql
```diff
@@ -2,11 +2,12 @@
 -- PostgreSQL database dump
 --
 
--- Dumped from database version 9.5.1
--- Dumped by pg_dump version 9.5.1
+-- Dumped from database version 9.6.5
+-- Dumped by pg_dump version 9.6.5
```
misaka: As @aliuk2012 has mentioned, I think, we need to make sure we're all on the same version of PostgreSQL, which is dictated by our deployed environments in AWS: 9.5.

PR #352, opened 2017-09-11
CT-1014 Allow user to view/edit teams he is a member of

stephenrichards:


app/policies/team_policy.rb
```diff
@@ -4,36 +4,64 @@ class TeamPolicy < ApplicationPolicy
...
 
   def can_add_new_responder?
     clear_failed_checks
-    check_user_is_a_manager
+    check_user_is_a_manager || check_user_is_member_of_team
   end
 
   def index?
     clear_failed_checks
-    check_user_is_a_manager
```
aliuk2012: If theres no checks is it worth even having this policy?

app/controllers/teams_controller.rb
```diff
@@ -3,8 +3,8 @@ class TeamsController < ApplicationController
 
 
   def index
-    @teams = BusinessGroup.all
-    authorize @teams.first
+    authorize Team
```
aliuk2012: Do we need to check user is authorised to view the index page? I would think that we are more concerned about the list of teams in the list and whether they can view those teams

spec/controllers/teams_controller_spec.rb
```diff
@@ -2,58 +2,154 @@
...
-          'commit' => 'Submit'}
-      }
+    describe 'GET index' do
+      it 'loads all teams the the repsonder is a member of' do
+        get :index
+        expect(assigns(:teams)).to eq [ business_unit, business_unit2 ]
+      end
 
-      before { sign_in manager }
+      it 'renders the index template' do
```
misaka: As we discussed in person, I feel a bit apprehensive about shoving all these specs into this controller spec after our experience with `cases_controller_spec.rb`, which currently stands at 1434 lines long, and that's with 3 of the verbs broken out into separate files. Can I propose that we:

1. change the organisation of the contexts so that instead of describing the verbs within the context of the user, we have the context of the user within the description of the verb. i.e. instead of this:
```
context 'manager' do
  describe 'GET show' do ...
  describe 'POST create' do ...
end

context 'responder' do
  describe 'GET show' do ...
  describe 'POST create' do ...
end
```
we do this:
```
describe 'GET show' do
  context 'manager' do ...
  context 'responder' do ...
end

describe 'POST create' do
  context 'manager' do ...
  context 'responder' do ...
end
```
I understand that this means that we won't be able to re-use quite as much setup, but we can mitigate that by using more fixtures (with `let` statements), or even better, my next point:

2. Move larger verbs, or maybe even every verb, out into a separate file. For examples have a look at the specs in `spec/controllers/cases_controller/*`

PR #354, opened 2017-09-12
Ct-968 adds bu group mailbox

beccapearce:


app/views/teams/_business_group_detail.html.slim
```diff
@@ -3,10 +3,6 @@ h2
     = t('team_leads.director_general')
   = @team.team_lead
 br
```
aliuk2012: Could you remove this tag its not needed

app/views/teams/_directorate_detail.html.slim
```diff
@@ -3,10 +3,6 @@ h2
     = t('team_leads.director')
   = @team.team_lead
 br
```
aliuk2012: Could you remove this tag its not needed

misaka:
@beccapearce I agree with @aliuk2012, about the lack of tests. I think that asking you to create a view test from scratch may be a bit much ... unless you really feel like doing it, could I ask you to please edit [CT-1049](https://dsdmoj.atlassian.net/browse/CT-1049) and add an item to the list there to add a spec for the show team view?

PR #357, opened 2017-09-14
CT-1077 Fix Extra buttons for Press and Private office

stephenrichards:
Unassigned press and private officers were seeing extra buttons.

This is as a result of permitted_events returning the wrong events.

This has been corrected, and a comprehensive test-suite for permitted events put in place


app/policies/case_policy.rb
```diff
@@ -269,17 +286,16 @@ def execute_request_amends?
   end
 
   def request_amends_from_pending_press_office_clearance_to_pending_dacu_clearance?
-    clear_failed_checks
```
misaka: This should probably be still in here.

app/policies/case_policy.rb
```diff
@@ -407,6 +423,14 @@ def responder_attachable?
     @user.in? BusinessUnit.press_office.approvers
   end
 
+  check :user_is_assigned_press_office_approver do
+    @user == self.case.approver_assignments.with_teams(BusinessUnit.press_office).first&.user
```
misaka: Is it worth it to be less dependent on there only being one press_officer assigned to the team? i.e., should we do this:

```
@user.in? self.case.approver_assignments.with_teams(BusinessUnit.press_office).map(&:user)
```

app/policies/case_policy.rb
```diff
@@ -407,6 +423,14 @@ def responder_attachable?
...
+  end
+
+  check :user_is_assigned_private_office_approver do
+    @user == self.case.approver_assignments.with_teams(BusinessUnit.private_office).first&.user
```
misaka: Same question as for `check :user_is_assigned_press_office_approver` above.

app/policies/case_policy.rb
```diff
@@ -407,6 +423,14 @@ def responder_attachable?
     @user.in? BusinessUnit.press_office.approvers
   end
 
+  check :user_is_assigned_press_office_approver do
```
misaka: This would read better as `check_user_is_assigned_press_officer_approver`. (`press_officer` vs `press_office`)

PR #362, opened 2017-09-19
CT-988 basic search

misaka:
https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&projectKey=CT&view=planning&selectedIssue=CT-988

Adds basic search via a new page in the global nav.

image


app/views/cases/index.html.slim
```diff
@@ -8,8 +8,11 @@
     .column-full.button-holder
       = form_tag search_cases_path, method: :get
         .form-group
+          = label 'query', 'Search for a case', class: 'form-label-bold'
```
aliuk2012: Label_tag and switch 'query' to :query

app/views/cases/index.html.slim
```diff
@@ -27,58 +27,59 @@
...
-              = t("state.#{kase.current_state}")
-            td aria-label="#{t('.who_its_with')}"
-              = kase.who_its_with
-            td aria-label="#{t('.message_notification')}"
-              - if kase.message_notification_visible?(current_user)
-                img { src=image_path('icons/message-bubble.svg')
-                      class="notification-speech-bubble"
-                      alt= "New notifications for case #{kase.number}" }
-    = paginate @cases
+- if current_page?(search_cases_path) && !@cases.nil?
```
aliuk2012: i.e what if the current page is open_cases or the other pages its shares shouldn’t we show the empty list?

misaka: Yes, that's botched, doing a fix.

PR #363, opened 2017-09-20
CT-1013 areas covered

aliuk2012:
Allows users to add/edit/delete business areas covered. It has been QA tested and has passed.

When a user is Add/Editing a business unit they should have to go through another step to add/edit areas covered. This is done using ajax.


app/controllers/teams_controller.rb
```diff
@@ -38,6 +47,7 @@ def new
     @team.team_lead = ''
     @team.parent_id = params[:parent_id].to_i
     @team_type = params[:team_type]
+    @action_copy = set_action_copy(true)
```
misaka: Can we call this `get_action_copy` or something similarly more descriptive? `set_action_copy` sounds like it should be reserved for setting `@action_copy` within the method, not by returning a value.

misaka: Also, how about using a keyword argument here so it's clearer what `true` means, e.g.:

```
@action_copy = get_action_copy(for_creation: true)
```

PR #366, opened 2017-09-21
CT-1096 Include environment and current user in Sentry report

stephenrichards:


app/controllers/application_controller.rb
```diff
@@ -31,4 +35,10 @@ def set_global_nav
       Settings.global_navigation,
     )
   end
+
+  # def set_raven_context
+  #   puts ">>>>>>>>>>>>>> SSETTING RAVEN CONTEXT #{__FILE__}:#{__LINE__} <<<<<<<<<<<<<<<<<\n"
+  #   Raven.user_context(id: current_user.id, email: current_user.email)
+  #   Raven.extra_context(host: ENV['ENV'] || 'Not set', myname: 'Stephen Richards')
+  # end
```
aliuk2012: Can we remove these?

PR #369, opened 2017-09-22
CT-1035 Show DS loads on cases resignments page

aliuk2012:


spec/decorators/user_decorator_spec.rb
```diff
@@ -0,0 +1,35 @@
+require 'rails_helper'
+
+describe CaseDecorator, type: :model do
```
stephenrichards: should be UserDecorator I think

PR #373, opened 2017-09-29
attempt to fix issues with user ids not being found after deactivation

beccapearce:


spec/features/admin/deactivating_users_spec.rb
```diff
@@ -21,6 +21,6 @@
     # A deactivated user cannot sign in
     cases_page.user_card.signout.click
     login_page.log_in(responder.email, responder.password)
-    expect(login_page.error_message).to have_content 'Invalid email or password'
+    expect(login_page.error_message).to have_content 'Your account is not activated yet.'
```
aliuk2012: I dont think we should be informing the user that their account is not activated. If I was a hacker attempting to login I would know that I have a valid username/password and could try another route. i.e speak to another staff member to get them to enable the account for me. (social engineering attacks)

aliuk2012: I know this is not very user friendly but it is the same reason why when they attempt to login using either a incorrect email or password  that we don't give any hints as to whether the user name or password were incorrect.

PR #375, opened 2017-10-02
CT-1046 Allow Disclosure to bypass Press and Private clearance

stephenrichards:


app/controllers/cases_controller.rb
```diff
@@ -218,10 +213,7 @@ def flag_for_clearance
                                     team: BusinessUnit.dacu_disclosure).call
   end
 
-  def approve_response
-    authorize @case, :can_approve_or_escalate_case?
```
misaka: Do we really want to be removing the `authorize` here?

app/views/cases/approve_response_interstitial.html.slim
```diff
@@ -0,0 +1,29 @@
+
+- content_for :heading
+  - if params[:mode] == 'clear'
+    = t('.clear')
+  - else
+    = t('.upload_and_clear')
```
misaka: Why is this `upload_and_clear`? So far as I can tell, "upload and clear" always renders `new_response_upload` and doesn't rely on this form, so when would this form be displayed for an uploaded response?

app/views/cases/approve_response_interstitial.html.slim
```diff
@@ -0,0 +1,29 @@
...
+.clearance-copy
+  p
+    = "You are clearing the response to: "
+    strong.strong-block
+      = @case.subject
+
+br
+
+= form_tag execute_response_approval_case_path(@case)
+  - if @case.state_machine.permitted_events(current_user.id).include?(:approve_and_bypass)
```
misaka: I like the use of `@case.allow_event?` ... if you're making changes anyway, might be worth using that here too.

app/services/bypass_params_manager.rb
```diff
@@ -0,0 +1,72 @@
...
+  #rubocop:disable Metrics/CyclomaticComplexity
+  def params_valid?
+    if present? && check_approval_requested? && check_message_present?
+      record_error 'Do not specify a reason for skipping further clearance if further clearance required.'
+    elsif present? && check_approval_not_requested? && check_message_blank?
+      record_error 'You must specify a reason for skipping further clearance'
+    else
+      true
+    end
+  end
```
misaka: Is it just me or would this read better as:

```
if present? && check_bypass_not_requested? && check_message_present?
  record_error 'Do not specify a reason for skipping further clearance if further clearance required.'
elsif present? && check_bypass_requested? && check_message_blank?
  record_error 'You must specify a reason for skipping further clearance'
else 
  true
end
```
Maybe it's just because this was my first encounter with it but since the whole point of this manager is to detect whether the user has requested a bypass, that's what I expected to be reading here.

app/views/cases/_bypass_approvals_form.html.slim
```diff
@@ -0,0 +1,17 @@
+#bypass-press-office-option
```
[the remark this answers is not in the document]
stephenrichards: Well this is a view page which specifically asks about press office, so I've left it hear.  But made it more generic everywhere else

PR #377, opened 2017-10-04
CT-1060 - Add performance percentage columns to R003

stephenrichards:
This PR adds three performance percentage reports to R003 -
Business Unit Performance Report as well as the name of the
team lead in each team.

This work is done by adding the ability to accept callback methods to the
Stats generator which allows the hierarchical rollup of statistics to be
accomplished, as well as addition of team lead roles.


app/services/stats/stats_collector.rb
```diff
@@ -73,6 +89,11 @@ def value(row, col)
 
     private
 
+    # def rollup_results(child_result_set, parent_team_id)
```
misaka: Eject me?

PR #379, opened 2017-10-05
CT-927 assignment case details

aliuk2012:


spec/controllers/assignments_controller_spec.rb
```diff
@@ -135,6 +135,30 @@
 
     before { sign_in responder }
 
+    describe 'GET edit' do
```
misaka: Can we move this into a section where the top-most context is the `describe 'GET edit' ...`? And we probably won't need to test different user types there since we don't authenticate in this controller action.

misaka:
I know it's not code you've changed, but we've noticed that `AssignmentsController#edit`, counter-intuitively, doesn't call `authorize` at the top of the function. It seems that we could move it up to make it a bit easier to read, and we'd benefit anyway from adding a `user_not_authorized` method that acts like the one in `UsersController` to redirect users that fail this particular authentication to the case details page, and not the case list view.

Worth doing, I think, only if it's a quick win and doesn't cause any other issues that need fixing.

app/controllers/assignments_controller.rb
```diff
@@ -204,8 +204,10 @@ def reassign_user_params
   end
 
   def set_case_and_assignment
-    if Assignment.exists?(id: params[:id])
+    if Case.exists?(id: params[:case_id])
```
misaka: This exposes another problem ... we are pulling up whatever case or assignment they want, but this means that users can manipulate the URL to discover cases that they don't have access too. Our default policy should be to only present cases and/or assignments that they have access to, I believe. This means:

1. The case should be retrieved using the Pundit policy scope: `Pundit.policy(current_user, Case).exists?(id: params[:case_id])`
2. We should additionally scope the assignments we access here to the case currently being viewed: `@case.assignments.exists?(id: params[:id])` (note, this is already how we retrieve @assignment).

I'm mentioning this here, but ... it's probably not worth fixing here as the issue existed before this change. We should create a ticket or make sure it's included in future tickets.

aliuk2012: I agree, this needs further work but that can be done another time. Thanks for your review 👍

PR #380, opened 2017-10-05
CT-762 add workflow support

misaka:
Groundwork for CT-762 Allow BMT to mark case as requiring clarification (TMM)

https://dsdmoj.atlassian.net/browse/CT-762

This change implements workflow support in policies and state machines. Please comment here on these changes, but as future work on CT-762 will be to actually implement a new policy and state machine, issues may be addressed there instead of here. Especially as the work there is likely to expose problems.

Additional comments and context in the commit message.


app/models/case.rb
```diff
@@ -21,10 +21,11 @@
 #  current_state        :string
 #  last_transitioned_at :datetime
 #  delivery_method      :enum
+#  workflow             :string
 #
 
 # Required in production with it's eager loading and cacheing of classes.
-require 'case_state_machine'
+# require 'case_state_machine'
```
stephenrichards: Do we need this now?

app/models/concerns/case_states.rb
```diff
@@ -1,12 +1,26 @@
...
-      self,
-      transition_class: CaseTransition,
-      association_name: :transitions
-    )
+    if @state_machine.nil?
+      state_machine_class_name =
+        if respond_to?(:workflow) && workflow.present?
+          raise NameError,
+                "State machine \"#{workflow}\" not found"
+        else
```
stephenrichards: is this raise just a temporary guard because nothing will have a non-standard workflow for the time being?
Perhaps we can consider populating every exising case with the standard workflow so that we don't have this code.

app/policies/case_policy.rb
```diff
@@ -1,313 +1,14 @@
...
-    check_user_is_a_manager
-  end
+    @case = kase
+    @user = user
 
-  def assign_to_new_team_from_drafting_to_awaiting_responder?
-    clear_failed_checks
-    check_user_is_a_manager
+    @workflow = policy_workflow_for_case(kase, user: user)
   end
```
stephenrichards: I think I'd rename @workflow and @policy, or @workflow_specific_policy or something to make it clearer that it is actually a workflow.

Like the delegtion in preference to inheritance

lib/event_transitions.rb
```diff
@@ -7,22 +7,39 @@ def initialize(machine, event_name, &block)
     instance_eval(&block)
   end
 
-  def transition(from: nil, to: nil, guard: nil, policy: nil)
+  def transition(from: nil, to: nil, guard: nil, policy: nil, workflow: nil)
     @from = to_s_or_nil(from)
```
stephenrichards: perhaps the workflow parameter could be called new_workflow, to remind us that it is only used when moving from one workflow to another.

lib/events.rb
```diff
@@ -38,6 +38,9 @@ def trigger!(event_name, metadata = {})
             + "with object: #{object} in state: #{object.current_state} " \
             + "and metadata: #{metadata}"
     end
+    if target_state.key? :workflow
+      object.update workflow: target_state[:workflow].to_s
+    end
```
stephenrichards: We've changed the workflow here - don't we have to reset the state machine as well?

misaka: This is handled by an `after_save` callback in `Case`, see [app/models/concerns/case_states.rb:5](https://github.com/ministryofjustice/correspondence_tool_staff/pull/380/files#diff-e3f8644866a7260c826920d478b45c33R5).

app/models/case.rb
misaka: No longer needed because we've moved the state machines into `app/state_machines`. This should also fix the auto-loading problems we were having.

app/models/concerns/case_states.rb
misaka: Good idea. I'll look to do this.

lib/event_transitions.rb
misaka: Good idea.

app/policies/case_policy.rb
misaka: :+1:

app/models/concerns/case_states.rb
misaka: So, I just remembered that we don't need to do this because when we add new case types, we'll actually actually use the `category` (or hopefully `type` if we rename it) column to set the default workflow. The workflow here will only be used if we aren't using the default workflow for that case type/category.

PR #383, opened 2017-10-09
CT-1034 Notify approvers when case is ready for clearance

aliuk2012:


app/mailers/assignment_mailer.rb
```diff
@@ -7,23 +7,48 @@ def new_assignment(assignment, recipient)
...
-        case_received_date: kase.received_date,
+        case_received_date: kase.received_date.strftime(Settings.default_date_format),
         case_subject:       kase.subject,
         case_link: edit_case_assignment_url(@assignment.case_id, @assignment.id)
     )
 
     mail(to: recipient)
   end
 
+  def ready_for_approver_review(assignment)
```
misaka: This class, `AssignmentMailer`, doesn't feel like the right place to put this. This mail isn't really about assignments ... I think we should move this into another file, although I'm not sure how best to organise these mailers. Maybe we should have a `ActionNotificationsMailer` which are all about notifying the user about actions to be taken within the system? Versus a mailer that may be a digest of the days activities ...

app/mailers/assignment_mailer.rb
```diff
@@ -7,23 +7,48 @@ def new_assignment(assignment, recipient)
...
+    RavenContextProvider.set_context
+
+    @assignment = assignment
```
misaka: We don't expose a view, we just pass in values to GovUK Notify, so we don't need to use `@assignment`, just plain `assignment is enough`.

app/services/case_approval_service.rb
```diff
@@ -60,5 +66,18 @@ def bypass_approval_for_team(team)
...
+  def notify_press_or_private
+    if @kase.current_state
+           .in?(%w( pending_press_office_clearance pending_private_office_clearance ))
 
+      current_info = CurrentTeamAndUserService.new(@kase)
+      assignment = @kase.approver_assignments
+                       .for_team(current_info.team)
+                       .first
+
+      AssignmentMailer
```
misaka: Why pull out `assignment`? It's not obvious why it's needed and I think it'd read better if you either just passed in `kase` and the mailer sorts out the approver to notify, or this service sorts out the approver and passes it through.

This goes along with changing the mailer to no just be about assignments.

app/services/case_approval_service.rb
```diff
@@ -34,16 +34,22 @@ def process_approval
...
                        .singular
         assignment.update!(approved: true)
+
         if @bypass_params.present? && @bypass_params.bypass_requested?
           bypass_press_and_private_approvals(assignment)
         else
           approve_and_progress_as_normal(assignment)
         end
+
+        assignment.reload
```
misaka: Still needed? It's not obvious what bit of the code needs this to be reloaded.

app/services/case_approval_service.rb
```diff
@@ -34,16 +34,22 @@ def process_approval
...
         @result = :ok
       end
     rescue Statesman::GuardFailedError
       @result = :error
     end
+
+    notify_press_or_private if @result == :ok
```
misaka: I know that this service already does this, but I don't think it should be talking about `press` or `private` ... it should be `next approver`, or something more generic, so that it'll make sense with other case types.

app/mailers/assignment_mailer.rb
aliuk2012: Good point 👍  and fixed

app/services/case_approval_service.rb
aliuk2012: Good point 👍  and fixed

app/services/case_approval_service.rb
aliuk2012: the assignment.kase.current_state is not updated unless assignment is reloaded. I need the case current state to be correct for the notify private method to work as expected. I agree it is not clear as to what or why this code is used for. For the time being I'll move it down to the notify_next_approver private method and run @kase.reload.

app/services/case_approval_service.rb
aliuk2012: So I did try to pass in just a case but the thing is we then have to lookup the approvers details for the mailer.  I also came across some weird errors relating to case being decorated and Stephen said that I should use kase.object when calling ready_for_approver_review but then I ran into other errors of 'NoMethodfound 'object' is undefined/

misaka: What about doing this:

```
AssignmentMailer.ready_for_approver_review(assignment.user, assignment.case).deliver_later
```

Again, it's just that it's not obvious why `ready_for_approver_review` is getting an assignment object, splitting it out like this is making intentions much clearer.

app/services/response_uploader_service.rb
```diff
@@ -30,6 +31,7 @@ def upload!
       attachments = @uploader.process_files(@uploaded_files, @type)
       transition_state(attachments)
       @result = :ok
+      notify_next_approver if @result == :ok && @action == 'upload-approve'
```
misaka: `@result` isn't ever going to be not `:ok` here ... it was just assigned. :)

misaka: Also, I liked the pattern in the `CaseApprovalService` where this notification is sent outside of rescue block so that it doesn't have an effect on the result of the action.

app/services/response_uploader_service.rb
```diff
@@ -102,4 +104,19 @@ def approve_and_progress_as_normal(filenames)
     )
   end
 
+  def notify_next_approver
+    if @case.current_state
+           .in?(%w( pending_press_office_clearance pending_private_office_clearance ))
```
misaka: This is just begging to cause a problem where we change this list of states in one location, either here or `CaseApprovalService`, and forget to change the other. Is there an easy way to centralise this that makes sense?

app/services/response_uploader_service.rb
aliuk2012: Sure, I agree I like the pattern in the ApprovalService but if I change it this there it has other impacts like how the controllers action processes the returned values and then theres all the tests so I dont think. its worth all the additional work for this ticket. We should standardise our approach and I think a separate ticket for it would be better suited

misaka: I don't see why this would be a big change ... all it would mean doing is moving the code in this method into a begin block, and then performing the `notify_next_approver` step outside of that block, leaving the `if @result == :ok && ...` check that's already here.

Although I agree that if it's a bigger change, this isn't the time to do it here,

app/services/response_uploader_service.rb
aliuk2012: While its already caused a problem and delay with getting this feature out by having two services "clearing" a case.

app/services/response_uploader_service.rb
```diff
@@ -25,20 +25,26 @@ def seed!(filepath)
...
-      attachments
+    begin
+      if @uploaded_files.blank?
+        @result = :blank
+        @attachments = []
+      else
+        @attachments = @uploader.process_files(@uploaded_files, @type)
+        transition_state(@attachments)
+        @result = :ok
+        @attachments
```
misaka: Unnecessary.

PR #386, opened 2017-10-10
CT-961 edit cases

stephenrichards:


app/controllers/cases_controller.rb
```diff
@@ -274,12 +284,12 @@ def create_foi_params
...
-    ).merge(category_id: Category.find_by(abbreviation: 'FOI').id)
+    ).merge(category_id: Category.foi.id)
   end
 
-  def parsed_edit_params
-    edit_params.delete_if { |_key, value| value == "" }
-  end
+  # def parsed_edit_params
+  #   edit_params.delete_if { |_key, value| value == "" }
+  # end
```
aliuk2012: Remove comments?

spec/controllers/cases_controller_spec.rb
```diff
@@ -117,37 +102,37 @@ def stub_current_case_finder_cases_with(result)
...
+    #   end
+    #
+    #   it 'assigns @case' do
+    #     expect(assigns(:case)).to eq(Case.first)
+    #   end
+    #
+    #   it 'renders the edit template' do
+    #     expect(response).to render_template(:edit)
+    #   end
+    # end
```
aliuk2012: shouldn't this be removed?

spec/controllers/cases_controller_spec.rb
```diff
@@ -117,37 +102,37 @@ def stub_current_case_finder_cases_with(result)
...
+    #     }
+    #
+    #     expect(Case.first.category.abbreviation).to eq 'GQ'
+    #   end
+    #
+    #   it 'does not overwrite entries with blanks (if the blank dropdown option is selected)' do
+    #     patch :update, params: { id: first_case, case: { category: '' } }
+    #     expect(Case.first.category.abbreviation).to eq 'FOI'
+    #   end
+    # end
```
aliuk2012: more comments

app/models/case.rb
```diff
@@ -28,6 +28,9 @@
 class Case < ApplicationRecord
   include Statesman::Adapters::ActiveRecordQueries
 
+  has_paper_trail only: %i{ name email received_date subject postal_address requester_type }
```
aliuk2012: There doesn't appear to be any tests for this.

stephenrichards: I don't think there is need to test this gem on every model we add it on.  Rebecca included tests when it was added to Team and User, so I think we can assume it works.

misaka: I read @aliuk2012's comment as being about _what_ attributes are being versioned, not that PaperTrail is working as advertised. However, the easiest way I can see to do this offhand is to check that when a case is changed, that the saved version has the expected attributes. Still, it would create a spec about what attributes get saved, which would be good.

app/controllers/cases_controller.rb
```diff
@@ -263,6 +272,7 @@ def process_closure_params
   end
 
   def create_foi_params
+    # category_param = ActionController::Parameters.new({ category_id: Category.foi.id })
```
misaka: No longer needed?

app/services/case_updater_service.rb
```diff
@@ -0,0 +1,27 @@
+class CaseUpdaterService
+
+  attr_reader :result
+
+  def initialize(user, kase, params)
+    @user = user
+    @team = @user.managing_teams.detect{ |t| t == BusinessUnit.dacu_bmt }
```
misaka: Why do this lookup? If `@user` has to be a part of `dacu_bmt`, then shouldn't this raise instead of setting `@team` to `nil`?

app/state_machines/cases/foi_state_machine.rb
```diff
@@ -190,6 +190,19 @@ def self.get_policy(user_id, object)
                authorize: true
   end
 
+  event :edit_case do
+    authorize :edit?
```
misaka: There is also a method `authorize_by_event_name` which I think would be good here. It takes no args and authenticate using `edit_case?` automatically, providing more consistency.

app/views/cases/_new_form_common.html.slim
```diff
@@ -0,0 +1,19 @@
+= f.text_field :name
+
+= f.email_field :email
+
+= f.text_area :postal_address, {rows: 4}
+
+= f.radio_button_fieldset :requester_type,
+        choices: [:academic_business_charity,
```
misaka: Surely we should grab this from the enum in the `Case` model? I see that `new.html.slim` does this too, but it feels like this should be centralised.

app/views/cases/show.html.slim
```diff
@@ -19,7 +19,8 @@
 = render partial: 'layouts/header'
 
 div id="case-#{@case.id}"
-  - if @permitted_events.any?
+
+  - if (@permitted_events - [:edit_case]).any?
```
misaka: Thought for later: feels like this is a candidate to be moved into `CaseDecorator`

app/models/case.rb
stephenrichards: Ok, so are we going to also write tests for Active Record, to check that find_by_full_name also  finds a record?

stephenrichards: Ok, well tests were justified.  We've found out that papertrail does not play nicely with acts_as_gov_uk_date.  Reify does not recreate the object correctly, but all the data is saved correctly in the version record.

app/services/case_updater_service.rb
stephenrichards: It will raise when trying to write the transition record, but I agree, there should be a better way of getting this.  ideas?

app/views/cases/_new_form_common.html.slim
stephenrichards: yep.  Case.requester_types.values

app/views/cases/show.html.slim
stephenrichards: Yes, I think we may well end up with different classes of permitted events which deserves an object of its own. PermittedEvents#action_buttons, PermittedEvents#links, etc.  need to think more about it and discuss with Al, - I know he has ideas.

app/services/case_updater_service.rb
misaka: Not elegant, but makes the intention clearer:

```
  raise("y u not dacu_bmt???") unless @team.present?
```

app/services/case_updater_service.rb
```diff
@@ -0,0 +1,27 @@
...
+    @team = kase.managing_team
+    @kase = kase
+    @params = params
+    @result = :error
+  end
+
+  def call
+    ActiveRecord::Base.transaction do
+      begin
+        @kase.state_machine.edit_case!(@user, @team)
```
aliuk2012: Shouldn't this be done after the update? otherwise we are saying the case has been editted when it hasn't and it might of failed due to validation errors etc.

app/controllers/cases_controller.rb
```diff
@@ -145,8 +150,16 @@ def upload_responses
...
+    authorize @case
+
+    service = CaseUpdaterService.new(current_user, @case, create_foi_params)
+    service.call
+    if service.result == :ok
       flash.now[:notice] = t('.case_updated')
+      set_permitted_events
+      @case = @case.decorate
+      @case_transitions = @case.transitions.case_history.order(id: :desc).decorate
       render :show
```
aliuk2012: This should be a redirect

app/services/case_updater_service.rb
```diff
@@ -13,9 +13,16 @@ def initialize(user, kase, params)
   def call
     ActiveRecord::Base.transaction do
       begin
-        @kase.state_machine.edit_case!(@user, @team)
-        if @kase.update(@params)
-          @result = :ok
+        @kase.assign_attributes(@params)
+        if @kase.changed?
+          if @kase.save!
```
misaka: `#save!` doesn't return true/false, it raises an exception. You should take out the `if ... else ... end` and replace it simply with:

```
        @kase.save!
        @kase.state_machine.edit_case!(@user, @team)
        @result = :ok 
```

spec/controllers/cases_controller_spec.rb
```diff
@@ -1445,18 +1445,21 @@ def do_upload_responses
       end
     end
 
-    context 'as a manager' do
+    fcontext 'as a manager' do
```
misaka: ^^^ You've probably noticed this by now ...

aliuk2012:
I've removed the fcontext

PR #389, opened 2017-10-18
CT-1108 Allow BMT to delete cases

aliuk2012:


app/controllers/cases_controller.rb
```diff
@@ -117,6 +119,23 @@ def edit
     render :edit
   end
 
+
+  def confirm_destroy
+    authorize @case, :can_delete_case?
+  end
+
+  def soft_delete_case
```
misaka: As discussed, this is probably better if we just use `destroy` here. I can see why you'd want to make it clear that this is not *actually* destroying the case, but it's confusing when everything else uses the wording `destroy` vs `soft_delete_case`.

app/controllers/cases_controller.rb
```diff
@@ -117,6 +119,23 @@ def edit
     render :edit
   end
 
+
+  def confirm_destroy
+    authorize @case, :can_delete_case?
```
misaka: I think we are better of doing something like:

```
def confirm_destroy
  authorize @case, :destroy?
  ...
end

# with the assumption that we're renaming #soft_delete_case to #destroy as per the comment below
def destroy
  authorize @case
  ...
end
```

This means renaming `can_delete_case?` to `destroy?` in `Cases::BasePolicy`, which we want to do to optimize for looking up policies in the policy file. This is because it's easier to lookup a policy if we can assume it's named after the controller action that uses it. This may mean that `confirm_destroy?` should have it's own policy which duplicates or calls `destroy?`, what do you think?

app/controllers/cases_controller.rb
```diff
@@ -117,6 +119,23 @@ def edit
...
+
+  def soft_delete_case
+    authorize @case, :can_delete_case?
+    service = CaseDeletionService.new(current_user, @case)
+    service.call
+    if service.result == :ok
+      flash[:notice] = "You have deleted case #{@case.number}."
+      redirect_to cases_path
+    else
+      render :confirm_destroy
```
misaka: Should this be a redirect? It might seem funny if the confirmation screen is showing up under the destroy URL, is all.

config/routes.rb
```diff
@@ -159,6 +159,8 @@
 
     get 'close', on: :member
     get 'closed' => 'cases#closed_cases', on: :collection
+    get 'confirm_destroy' => 'cases#confirm_destroy', on: :member
+    delete 'destroy' => 'cases#destroy', on: :member, as: :destroy
```
misaka: Don't think we need the `as: :destroy` on here anymore.

db/migrate/20171013134445_add_deleted_to_cases.rb
```diff
@@ -0,0 +1,5 @@
+class AddDeletedToCases < ActiveRecord::Migration[5.0]
+  def change
+    add_column :cases, :deleted?, :boolean, default: false
```
misaka: Since we have a default scope on this field, I think we want to have it indexed.

spec/models/case_spec.rb
```diff
@@ -151,6 +151,12 @@
     end
   end
 
+  describe 'default_scope' do
+    it "applies a default scope to exclude deleted cases" do
+      expect(Case.all.to_sql).to eq Case.unscoped.where( deleted?: false).to_sql
```
misaka: Using `to_sql` here makes this test a bit brittle ... if we have to add something else to the default scope, like ordering, or another filter, then this will fail. AND we won't be able to only test that the scope excludes deleted cases since we're testing the entire generated SQL command. I'd recommend (i.e. like to see but not block the PR for :) ) that we create a deleted case and a non-deleted case that are otherwise identical, and just test that only the non-deleted one shows up.

spec/state_machines/cases/foi_state_machine_spec.rb
```diff
@@ -273,6 +273,50 @@ def event(event_name)
...
+                  .to(:pending_press_office_clearance)
+                  .checking_policy(:edit_case?) }
+    it { should transition_from(:pending_private_office_clearance)
+                  .to(:pending_private_office_clearance)
+                  .checking_policy(:edit_case?) }
+    it { should transition_from(:responded).to(:responded)
+                  .checking_policy(:edit_case?) }
+  end
+
+  events :delete_case do
```
misaka: You know what could really give us a net benefit here? Is if we iterated through each of the states available in the state machine and ensured that a transition existed for it. That would mean that any new states we add would be automatically tested to ensure we don't forget to add a transition here for it.

spec/models/case_spec.rb
aliuk2012: if we did change the default scope and this failed then it would have done its job ;-)
If we created another filter/scope then it shouldn't cause this test to fail as we would have to call them specifically on Case for it to fail

spec/state_machines/cases/foi_state_machine_spec.rb
aliuk2012: I agree that it would benefit us to have a any? state so we did not have to hard code each state but given time restraints and we have other existing code that repeats this I think we should create a ticket that addresses this issue.

app/controllers/cases_controller.rb
aliuk2012: no I dont think its an issue as you will see that  we  dont redirect if there are validation errors. eg. 
Creating a new case 
https://github.com/ministryofjustice/correspondence_tool_staff/blob/develop/app/controllers/cases_controller.rb#L87

Updating an existing case
https://github.com/ministryofjustice/correspondence_tool_staff/blob/develop/app/controllers/cases_controller.rb#L170

spec/state_machines/cases/foi_state_machine_spec.rb
misaka: :+1:

app/controllers/cases_controller.rb
misaka: 👍

spec/models/case_spec.rb
misaka: Although I disagree that we want this test to break if the default scope changes (the spec says it excludes deleted cases, not that it should be exactly that SQL), we can worry about this later / next time we implement a default scope. 👍

PR #392, opened 2017-10-20
CT-762 new workflows

misaka:
This brings in some key, but incomplete, changes that will be useful when we
need to implement other case types. These changes were originally designed
around implementing a workflow for TMMs, but the requirements for that feature
changed suddenly (once the stakeholders saw what we implemented *sigh*) so
this is the result of extracting the non-TMM specific code from the changeset.

The key changes are that were being worked on were:

- Specify FOI case category when creating a generic Case object. For example,
  for use with Pundit or when presenting the "Add case" button.
- When triggering state machine events allow the state machine to decide
  whether the acting team is the responding_team, managing_team, etc.
  Different workflows may have different rules to apply here for TMMs the
  acting user was in the managing team, for example.
- Don't rely on @permitted_events to decide which action buttons show on the
  case details page, use the case policies instead. Case policies then
  authorize the user _and_ check that the state machine event is triggerable.
- Focus event "policies" on checking the state of the case, not who is
  performing the action, that is the domain of the controller.


app/controllers/cases_controller.rb
```diff
@@ -377,6 +375,20 @@ def get_flash_errors_for_case(kase)
...
   end
+
+  def authorize_upload_response_for_action(kase, action)
+    case action
+    when nil, 'upload'    then authorize kase, 'upload_responses?'
+    when 'upload-flagged' then authorize kase, 'upload_responses_for_flagged?'
+    when 'upload-approve' then authorize kase, 'upload_responses_for_approve?'
+    when 'upload-redraft' then authorize kase, 'upload_responses_for_redraft?'
+    # else authorize kase, 'upload_responses?'
+    end
```
stephenrichards: Presumably this commented out line can be removed.

PR #394, opened 2017-10-20
CT-1045 adds disclosure removal service

beccapearce:
Not sure I've done this in the best way so do be ruthless!

Also looking back over it I'm not sure if I've dealt with the message for the state machine particularly well
(A message is required to remove clearance after a case is taken on but not when it is deescalated before it has been taken on).


app/controllers/cases_controller.rb
```diff
@@ -352,5 +373,8 @@ def get_flash_errors_for_case(kase)
       end
     end
   end
+  def redirect?(kase)
```
aliuk2012: do we need this empty method?

app/services/case_unflag_for_clearance_service.rb
```diff
@@ -1,31 +1,35 @@
...
-    else
-      @result = :not_flagged
-      false
-    end
+  def flagged_for_press_or_private?
+    @case.flagged_for_private_office_clearance? || @case.flagged_for_press_office_clearance?
+  end
+
+  def eligible?
+    @case.requires_clearance? && !flagged_for_press_or_private?
```
aliuk2012: dont we have this as a policy? it not i think maybe we should add it

spec/controllers/cases_controller_spec.rb
```diff
@@ -1122,10 +1122,57 @@ def do_upload_responses
     end
   end
 
+  describe 'PATCH unflag_taken_on_case_for_clearance' do
```
aliuk2012: Could we put this describe into its own seperate file? The reason is that as you can see that this case controller spec is ridiculously long and we started breaking out  actions into their own file which you can find going to `spec/controllers/cases/`

aliuk2012: Also it you look at some of the existing spec you should see ways that we test authorisation without going through all the user types

app/controllers/cases_controller.rb
```diff
@@ -249,6 +266,10 @@ def flag_for_clearance
                                     team: BusinessUnit.dacu_disclosure).call
   end
 
+  def remove_clearance
+    @case = Case.find(params[:id]).decorate
```
aliuk2012: What is the purpose of this action? You have added it to before_Action set_case and that calls set_case. Set_case finds the case based on the id param ad decorates it as you have done here (eg. https://github.com/ministryofjustice/correspondence_tool_staff/blob/develop/app/controllers/cases_controller.rb#L324)

app/services/case_unflag_for_clearance_service.rb
```diff
@@ -1,31 +1,22 @@
...
   end
 
   def call
-    return @result unless validate_case_is_flagged
-
-    @case.state_machine.unflag_for_clearance!(@user, @case.managing_team, @team)
-    @case.approver_assignments.with_teams(@team).destroy_all
-
-    @result = :ok
+      @case.state_machine.unflag_for_clearance!(@user, @case.managing_team, @team, @message)
```
aliuk2012: Can you wrap a transaction around it.

PR #395, opened 2017-10-20
CT-602 pre-check if teams exist for smoke tests

aliuk2012:
FactoryGirls find_or_create method is very strict on what to find
and what I have found is that the factories create a slighly different
version(as in the attributes have different attribute values) to what
is used in the CTS scripts. So as a hack i've wrapped an if condition
that checks if any BMT or DS team are present before running FactoryGirl
find_or_create method.


aliuk2012:
I changed it to a straight create

PR #402, opened 2017-10-24
Ct1110 case closure

stephenrichards:


config/locales/en.yml
```diff
@@ -155,6 +155,7 @@ en:
         team_id: Assign to
         user_id: Choose a new team member for this case
       case:
+        info_held_status_abbreviation: Was the information held?
         requester_type: Type of requester
         refusal_reason_name: What was the reason for refusal?
```
aliuk2012: Needs to be updated as mentioned in the QA results

app/models/case_closure/exemption.rb
```diff
@@ -50,13 +54,17 @@ class Exemption < Metadatum
     scope :absolute, -> { where(subtype: 'absolute') }
     scope :qualified, -> { where(subtype: 'qualified') }
 
+    def self.absolute_for_partly_refused
+      absolute.where.not(abbreviation: 'cost')
+    end
+
```
misaka: Why not a scope?

app/models/case_closure/exemption.rb
```diff
@@ -50,13 +54,17 @@ class Exemption < Metadatum
...
 
-    def self.method_missing(method)
+    def self.method_missing(method, *args)
```
misaka: It's a good idea to also define `respond_to_missing?`, otherwise `#respond_to?` and `#method` are broken:

```
irb(main):007:0> cc = CaseClosure::Exemption.s21
=> #<CaseClosure::Exemption ...>
irb(main):008:0> cc = CaseClosure::Exemption.respond_to? :s21
=> false
irb(main):010:0> CaseClosure::Exemption.method :s21
NameError: undefined method `s21' for class `#<Class:0x007fcf83e2c728>'
```

https://robots.thoughtbot.com/always-define-respond-to-missing-when-overriding

app/models/case_closure/info_held_status.rb
```diff
@@ -0,0 +1,60 @@
...
+#  active                  :boolean          default(TRUE)
+#  required_for_refused    :boolean          default(FALSE)
+#  required_for_ncnd       :boolean          default(FALSE)
+#
+
+module CaseClosure
+  class InfoHeldStatus < Metadatum
+
+    def self.held
+      where(abbreviation: 'held').first
```
misaka: Should this and the ones below use `.singular` instead of `.first`?

spec/validators/closed_case_validator_spec.rb
```diff
@@ -2,150 +2,402 @@
 
 describe 'ClosedCaseValidator' do
 
+  before(:all) do
+    require File.join(Rails.root, 'db', 'seeders', 'case_closure_metadata_seeder')
+    CaseClosure::MetadataSeeder.seed!
+    # CaseClosure::MetadataSeeder.implement_oct_2017_changes
```
misaka: I guess we don't need this anymore.

app/models/case_closure/exemption.rb
```diff
@@ -68,6 +64,10 @@ def self.method_missing(method, *args)
       end
     end
 
+    def self.respond_to?(method, include_private = false)
```
misaka: Should be `respond_to_missing?` :) This is a special version of `respond_to?` that's meant to be used in coordination with `method_missing`.

PR #412, opened 2017-10-31
CT-1191 Add a Trigger marker to case details

aliuk2012:


spec/decorators/case_decorator_spec.rb
```diff
@@ -125,6 +125,24 @@
     end
   end
 
+  describe '#trigger_case_marker' do
+    context 'unflagged case' do
+      it 'returns space' do
```
misaka: Why does it need to return a space?

aliuk2012: the table cell needs it.  We do this for the draft deadline too https://github.com/ministryofjustice/correspondence_tool_staff/blob/develop/app/decorators/case_decorator.rb#L35

spec/decorators/case_decorator_spec.rb
```diff
@@ -125,6 +125,24 @@
...
+        unflagged_case = create(:case).decorate
+        expect(unflagged_case.trigger_case_marker).to eq ' '
+      end
+    end
+
+    context 'flagged case' do
+      it 'returns the internal deadline' do
```
misaka: Wrong spec here I think, but not worth blocking the PR for it.

aliuk2012: doh, copy and paste

PR #415, opened 2017-11-01
Ct 1185 redraft and message notifications

beccapearce:


spec/mailers/action_notifications_mailer_spec.rb
```diff
@@ -113,5 +108,51 @@
...
+           case_current_state: 'ready to send',
+           case_number: approved_case.number,
+           case_abbr: 'FOI',
+           case_name: 'Fyodor Ognievich Ilichion',
+           case_received_date: 10.business_days.ago.to_date.strftime(Settings.default_date_format),
+           case_subject: 'The anatomy of man',
+           case_link: case_url(approved_case.id),
+           case_draft_deadline: approved_case.internal_deadline.strftime(Settings.default_date_format),
+           case_external_deadline: approved_case.external_deadline.strftime(Settings.default_date_format)
+           })
```
misaka: As per our convo ... this is already tested in the general spec so it's a little distracting to see it here ... it'd be worth doing the same as the other contexts here and just test that the `email_subject` is set.

spec/mailers/action_notifications_mailer_spec.rb
```diff
@@ -113,5 +108,51 @@
...
       expect(mail.to).to include assignment.user.email
     end
+
+    context 'ready to send' do
+      let(:mail) { described_class.notify_information_officers(approved_case, 'Ready to send')}
+      it 'sets the template' do
+        expect(mail.govuk_notify_template)
+          .to eq '46dc4848-5ad7-4772-9de4-dd6b6f558e5b'
+      end
+    end
```
misaka: Personally, I'd like to see at least one line of whitespace between describes, contexts and specs.

